Issue #16, opened 2026-02-06
Single wallet  can register multiple tokens


ruhil6789:
Now We are using the adminpkh  as a single identity  while registering the token.

nemo83:
This is not an issue of cip113, but of the substandard.

PR #17, opened 2026-02-06
 same wallet register multiple token using asset name

ruhil6789:
Issuer admin contract that allows transfers signed by a permitted credential
AND enforces asset name consistency to prevent policy ID collisions

 This validator solves the problem where multiple tokens from the same admin
 would have the same policy ID. By including the asset_name as a parameter,
 each token gets a unique script hash, resulting in unique policy IDs.

 **Mainnet Production Notes:**
 - This validator is parameterized with both admin credential AND asset name
 - The asset name parameter ensures each token has a unique policy ID
 - Validation checks that minted tokens match the expected asset name
 - Seize operations (no minting) bypass asset name validation
   ***Related Issue***
   #16 
 @param permitted_cred - The credential authorized to control this token (admin)
 @param asset_name - The token name this contract is bound to (prevents reuse)

### DIFF
--- a/src/programmable-tokens-onchain-aiken/lib/utils.ak
+++ b/src/programmable-tokens-onchain-aiken/lib/utils.ak
@@ -189,3 +189,19 @@ pub fn is_programmable_token_registration(
 
   has_minting? && cs_matches?
 }
+
+/// Check if a list of integers is strictly increasing (each element > previous)
+/// This guarantees both uniqueness and sorted order
+/// Used for security validation of index lists in ThirdPartyAct
+pub fn is_strictly_increasing(idxs: List<Int>) -> Bool {
+  when idxs is {
+    [] -> True
+    [_] -> True
+    [first, second, ..rest] ->
+      if first < second {
+        is_strictly_increasing([second, ..rest])
+      } else {
+        False
+      }
+  }
+}
--- a/src/programmable-tokens-onchain-aiken/validators/programmable_logic_global.ak
+++ b/src/programmable-tokens-onchain-aiken/validators/programmable_logic_global.ak
@@ -14,7 +14,7 @@ use types.{
 }
 use utils.{
   bytearray_lt, elem_at, expect_inline_datum, expect_value_contains_v3,
-  has_currency_symbol, is_script_invoked, is_signed_by,
+  has_currency_symbol, is_script_invoked, is_signed_by, is_strictly_increasing,
 }
 
 validator programmable_logic_global(protocol_params_cs: PolicyId) {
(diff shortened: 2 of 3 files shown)


ruhil6789:
Hi @nemo83 , Please take a look at the aiken script which help us  adding asset name as the second identity to register multiple tokens.

src/substandards/freeze-and-seize/validators/example_transfer_logic.ak
```diff
@@ -4,35 +4,114 @@ use aiken/collection/list
...
+        // Minting is happening → validate asset name matches
         list.any(
-          self.withdrawals,
-          fn(wdrl) {
-            let Pair(cred, _amount) = wdrl
-            cred == script_cred
+          flatten(self.mint),
+          fn(token) {
+            let (_policy_id, token_name, _quantity) = token
+            token_name == asset_name
```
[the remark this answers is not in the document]
nemo83: what do you mean by binding?

I think we need to parameterised the issuance part of the contract w/ 

1. issuance_cbor_hex_cs
2. prog base token hash

Then pass in the redeemer the hash of the withdraw contract which will then

1. calculate policy id via hashing issuance prefix, progr bas + hahd in redeemer + issuance postfix) 
2. check tehre is mint of policy at 1 + aaseet name , where asset name is passed as well as contract param.
3. ensure the hash passed in the redeemer is actually executed in the tx.

This could work, I think.

nemo83 (changes requested):
plese remove all the `*.class` files which ended up in the PR.

src/programmable-tokens-offchain-java/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Tue Jan 07 15:36:18 GMT 2025
+#Fri Feb 06 16:34:33 IST 2026
 distributionBase=GRADLE_USER_HOME
-distributionPath=wrapper/dists
 distributionUrl=https\://services.gradle.org/distributions/gradle-8.8-bin.zip
-zipStoreBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
 zipStorePath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
```
nemo83: please rollback

nemo83 (changes requested):
Hi, thanks for the PR, while I review the code, can you please continue removing unnecessary source code:

* there is a `bin` folder in the java project. please remove it
* remove all log files that are part of the 
* rollback changes to the gradle-wrapper.propeties
* remove pid file(s)

src/programmable-tokens-offchain-java/src/main/java/org/cardanofoundation/cip113/service/UtxoProvider.java
```diff
@@ -68,20 +95,29 @@ public List<Utxo> findUtxos(String address) {
                 return utxos;
             }
 
+>>>>>>> upstream/main
```
nemo83: more unresolved conflict

ruhil6789:
@nemo83  I have only deployed  the Aiken script.

PR #31, opened 2026-03-04
Rework, fixes and optimise 3rd party action

KtorZ:
I went on reviewing and reworking the 3rd party action upon request from @nemo83.

  ### Fixes

  While working through the code and cross-comparing with the Plutarch
  reference implementation, I've found several issues (which I fixed
  along the way):

  1. The "indices" in the redeemer have been misunderstood / misuses as indices, whereas they are used as "skip offsets" in the Plutarch implementation. This mostly only matters when there are more than 1 input being validated; Note that the current situation is still not ideal as it requires some patching for the first indice. It would be better if indices assumed that input were consumed along the way.

     Note that I had to adjust a couple of tests due to that. I've tried to simply alter them to match that behaviour, instead of just removing them.

  2. The previous implementation wasn't computing deltas for each input/output pair, but was considering the full input amount. Although this differs from the Plutarch implementation which only collects delta, I couldn't convince myself that it was wrong? If anything, it makes the check *stronger* since we are now asserting that ALL input tokens must still live in programmable tokens (or be burned), instead of asserting that only the difference is. It provides a nice optimisations, so I kept it, with a note.

  3. The Plutarch implementation allows indices pointing at script-locked inputs that are locked by an arbitrary script credential. I don't really understand why (haven't given much thought, but least I can say is that isn't *obviously correct*). Yet, I've aligned the new implementation with this, since the old would halt in this situation.

  ### Optimisations

  I've been using similar strategies as for the transfer, with an extra care about traversing inputs / outputs only once; making sure that enough context is accumulated in a single pass to perform all the necessary checks.

  ### Quick Benchs

  Note: I've translated the following preview transaction as an Aiken baseline bench: https://preview.cexplorer.io/tx/d29a977a55ed54ef0f5c9d5c15f85a63e915349890f1952934809070430ef33a

  From it, I've created a second baseline a bit more interesting with 3rd-party actions from multiple inputs, with additional unrelated tokens in inputs and outputs and with some burning, to also analyze how the code 'scales'.

  A summary of the executions can be found here:

  ##### Baseline 1

  | Version     | mem     | cpu     |
  | ---         | ---     | ---     |
  | Plutarch    | 228.09K |  85.23M |
  | Aiken (old) | 308.97K | 112.09M |
  | Aiken (new) | 231.35K |  84.61M |

  ##### Baseline 2

  | Version     | mem     | cpu     |
  | ---         | ---     | ---     |
  | Plutarch    | N/A     | N/A     |
  | Aiken (old) | 552.90K | 196.99M |
  | Aiken (new) | 356.49K | 129.79M |


KtorZ:
@nemo83 added a couple of tests regarding that last TODO + simplified slightly the 3rd party check by requiring that all provided indices in the 3rd party act redeemer to point at programmable inputs. As discussed, we do traverse all inputs anyway so there's no need to allow this (and thus, it's better to forbid it entirely).

PR #34, opened 2026-03-13
Split and tidy code in preparation for audit and testing.

KtorZ:
## Overview

Near-final PR to prepare the code for audit and external reviews. No real semantic changes in this PR, but code re-organisation following a simple rule: one 'type' per module. From experience, this leads to cleaner codebases, greater code reuse, and easier testing. It also makes the overall code a bit more breathable.

### Disclaimer

Besides, I also took some time to compare this with the latest Plutarch's version (from [input-output-hk/wsc-poc#107](https://github.com/input-output-hk/wsc-poc/pull/107)). Having an exact comparison has been challenging because both implementations have significantly reduced execution costs and are now competing in the thousands of lovelaces range.

#### Regarding costs

For example, a simple transfer of programmable tokens requires, give or take, about 200K mem units and 75M cpu units. That's about 1.2% and 0.75% of a transaction's total execution budget. So, effectively, a 1% difference between the two implementations corresponds to an impact of about 0.01% on the overall transaction budget. Said differently, a 1% difference across both mem and cpu between these two implementations results in a difference of execution cost of around 200 lovelaces. Surely, this piles up over time and across millions of transactions.

While important, execution units are also to be balanced with script size. And here, every byte difference between two scripts results in *at least* 15 lovelace of overhead; which may compensate or compound with execution units.

#### Finding a common ground for fair comparison

It was a bit tricky to get the right numbers for comparing the two implementations because Plutarch's benchmarks are set up differently from Aiken's. In particular:

1. The Aiken benchmarks always include the cost of running both the base `spend` validator, alongside the global `withdraw` one. Whereas the Plutarch bench is more validator-focused and only measures the cost of the global `withdraw` validator.

2. Similarly, Aiken's benchmarks all assume the transfer logic to be script-based (thus requiring an extra withdrawal script); whereas the Plutarch benchmarks all assume a verification-key-based transfer logic (consequently looking into transaction extra signatories).

3. There were also subtle differences, such as extra noise inputs and outputs in one benchmark, or in the ordering of inputs and/or registry nodes.

These may all sound superficial, but when comparing implementations that account for less than 1% of the total budget, I believe it is important to have benches as close as possible to one another. So I did my best to adjust the Aiken benchmarks locally to match these few points (although this hasn't been committed, I still find it more meaningful to look at the more plausible scenarios where the spend script is executed alongside the withdraw script, and where programmable tokens are script-owned). I have also extended the Plutarch benchmarks to cover more scenarios to ensure a fairer comparison across the board. Finally, I ported the `mixed` bench from Plutarch to Aiken's, which adds a nice "complex" baseline that mixes a bit of everything. So, the final bench set includes the following scenarios:

- **Transfer (simple):** original 'baseline' scenario, one input with programmable tokens split between two outputs.
- **Transfer (1 mint):** the 'simple' baseline augmented with an extra minting operation where the minted tokens are sent towards one of the two outputs.
- **Transfer (4 tokens):** a scenario where a single programmable input holding 4 tokens under the same policy is spent and sent to a single output.
- **Transfer (5 outputs):** a case where a single programmable input holding a single token is sent to 5 different outputs.
- **Transfer (15 inputs):** an aggregation of 15 programmable inputs holding a single token kind, sent to a single output.
- **Transfer (mixed):** a mixed scenario with several inputs, several tokens, from several policies, to several outputs. Inputs include, in particular, non-programmable tokens too.
- **Seize (1):** a simple 3rd-party action where all tokens from a single input are seized.
- **Seize (10):** like seize (1), but across 10 inputs, with 10 corresponding outputs and an additional 3rd party output.

So, these are not _perfect_ as they still do not test good mixes of mint, burn, policies, and non-programmable inputs/outputs, but they should already give a clearer picture.

#### Results & comparison

Here are the latest results for the presented benchmarks, adjusted according to the points (1), (2), and (3) mentioned above. Some subtle differences may still be found on either side.

| VERSION         | SCRIPT\_SIZE (global) | SCRIPT\_SIZE (base) |
| :------         | --------------------: | ------------------: |
| Plutarch        | 4382 bytes            | 140 bytes           |
| Aiken (pre-#31) | 3367 bytes            | 139 bytes           |
| Aiken (latest)  | 2834 bytes            | 139 bytes           |

| BENCH                | VERSION         |     MEM |     CPU |   ΔMEM |   ΔCPU |
| :------------------- | :-------------- | ------: | ------: | -----: | -----: |
| Transfer (simple)    | Plutarch        | 210.80K |  72.61M |     0% |     0% |
| Transfer (1 mint)    | Plutarch        | 252.06K |  88.37M |     0% |     0% |
| Transfer (4 tokens)  | Plutarch        | 275.50K |  89.63M |     0% |     0% |
| Transfer (5 outputs) | Plutarch        | 318.66K | 110.47M |     0% |     0% |
| Transfer (15 inputs) | Plutarch        | 969.11K | 317.39M |     0% |     0% |
| Transfer (mixed)     | Plutarch        | 446.30K | 147.02M |     0% |     0% |
| Seize 1              | Plutarch        | 213.66K |  69.99M |     0% |     0% |
| Seize 10             | Plutarch        | 789.44K | 289.99M |     0% |     0% |
| -------------------- | --------------- | ------- | ------- | ------ | ------ |
| Transfer (simple)    | Aiken (pre-#31) | 229.06K |  78.59M |  +8.7% |  +8.2% |
| Transfer (1 mint)    | Aiken (pre-#31) | 219.77K |  74.28M | -12.8% | -15.9% |
| Transfer (4 tokens)  | Aiken (pre-#31) | 201.27K |  68.01M | -26.9% | -24.1% |
| Transfer (5 outputs) | Aiken (pre-#31) | 417.19K | 138.23M | +30.9% | +25.1% |
| Transfer (15 inputs) | Aiken (pre-#31) |   1.08M | 357.27M | +11.4% | +12.6% |
| Transfer (mixed)     | Aiken (pre-#31) | 478.48K | 156.21M |  +7.2% |  +6.3% |
| Seize 1              | Aiken (pre-#31) | 195.48K |  70.36M |  -8.5% |  +0.5% |
| Seize 10             | Aiken (pre-#31) | 815.97K | 315.03M |  +3.4% |  +8.6% |
| -------------------- | --------------- | ------- | ------- | ------ | ------ |
| Transfer (simple)    | Aiken (latest)  | 193.60K |  68.95M |  -8.2% |  -5.0% |
| Transfer (1 mint)    | Aiken (latest)  | 179.33K |  62.73M | -28.9% | -29.0% |
| Transfer (4 tokens)  | Aiken (latest)  | 176.80K |  62.90M | -35.8% | -29.8% |
| Transfer (5 outputs) | Aiken (latest)  | 318.88K | 110.51M |  +0.1% |   0.0% |
| Transfer (15 inputs) | Aiken (latest)  | 899.85K | 306.28M |  -7.1% |  -3.5% |
| Transfer (mixed)     | Aiken (latest)  | 307.89K | 107.93M | -31.0% | -26.6% |
| Seize 1              | Aiken (latest)  | 180.17K |  63.60M | -15.7% |  -9.1% |
| Seize 10             | Aiken (latest)  | 795.61K | 307.57M |  +0.8% |  +6.1% |

#### Extrapolations

The Aiken implementation also provides scaling benchmarks for inputs, tokens, policies, and outputs. They show that the growth of execution units is linear or quasi-linear on those dimensions, allowing us to make a few projections regarding the maximum in any of these dimensions (and assuming that they are somewhat independent, even if they aren't fully), considering the current Mainnet execution limits for `mem=16500000` & `cpu=10000000000`:

| DIMENSION | PROJECTED MAXIMUM |
| --------- | ----------------- |
| Inputs    | 338               |
| Outputs   | 391               |
| Tokens    | 2007              |
| Policies  | 147               |

## Conclusion

Overall, the latest round of optimisation we did on the Aiken codebase has allowed us to both reduce script size and significantly lower execution costs. Plutarch's implementations still appear to perform slightly better overall for scenarios with many outputs and for large third-party actions. In the Aiken implementation,we have indeed placed emphasis on code reuse to reduce overall script size, leading to some sacrifices in 3rd-party logic. Given that 3rd-party actions should likely be _less_ frequent than the transfer logic, I believe this is an adequate trade-off.

There's likely not much we can gain from performances now unless there are major architectural changes in the design.


nemo83 (approved):
Thanks for the great PR, splitting code into modules greatly increases readibility of the codebase. 

Thanks for fixing the `is_updated_directory_node` workflow.

Ran some AI assisted stuff and spotted areas to improve code coverage as per below and might try to implement some myself.

### Suggestions for follow-up (non-blocking)

1. **`assets.split_at()`** — no dedicated unit tests currently; this function is used in registry/transfer logic and would benefit from edge-case testing (split at 0, at length, beyond length)
2. **`list.{get_or_fail, drop_or_fail, skip_assert_or_fail}`** — no unit tests; the unrolled recursion pattern is clever but worth verifying independently
3. **Module-level docstrings** — the new `lib/` files would benefit from brief module-level documentation describing their purpose and invariants
4. **No malformed datum deserialization tests** — consider adding tests for graceful handling of unexpected datum formats

PR #39, opened 2026-03-18
simplify 3rd party action by removing inputs / outputs indices.

KtorZ:
## Why ?

  The input indices were not really relevant in the current
  implementation, because we would still go over all inputs. This
  slightly differs from Plutarch's which would allow skip through
  non-script inputs, and then perform an extra check on the redeemers to
  ensure that no inputs were missed.

  Furthermore, these indices are annoying to compute for whomever is
  building the transaction, since inputs on Cardano are alphabetically
  sorted (so one must construct the transaction before they can
  construct the transaction :grimacing:).

  ## Consequences

  Overall, the worse case scenario is when one is trying to collect from
  many programmable inputs at once where all selected inputs are
  programmable *anyway*. The other most obvious use case is when seizing
  one particular UTxO, where inputs are very likely limited also. So in
  both case, the added value of those skip indices is unclear.
 
  Note that I have **also** removed the output skip indices, and required 
  seized outputs to always appear *first* in the transaction. This is perhaps 
  debatable as it *could* conflict with another script-locked input having a 
  similar requirement; so that clearly effectively reduces the composability 
  of the validator. Yet, in the case of the 3rd party action, this seems like 
  a reasonable trade-off?

  Besides, I've added a new baseline benchmark for the 3rd party action
  collecting from 150 inputs at once; which is about the maximum which
  can happen on-chain anyway due to the transaction max size. So long as
  execution cost can remain below the transaction maximum here, we are
  likely fine.

  Interestingly, the benchmark comparison between the two versions show
  that the 'optimised' version using indices is not only slightly worse
  in all cases, but also leads to larger scripts (by ~300 bytes) because
  more logic is needed to handle them. So in summary, indices are:

  1. leading to worse cost
  2. increasing script size
  3. more annoying to use off-chain

  So removing them seems like a no-brainer.

  Either way, both version are well below the transaction max execution
  costs when maxing out the inputs / outputs for single tokens. It could
  be interesting to benchmark with a mix of tokens too although I do not
  expect a significant difference between the two approaches (the
  benchmark is simply intersting on its own!).

  ### With indices

  Script size (global logic): 2834 bytes

  | bench                | mem      | cpu      |
  | -------------------- | -------: | -------: |
  | baseline_3rd_party_1 | 203.78 K |  77.30 M |
  | baseline_3rd_party_2 | 333.70 K | 123.94 M |
  | baseline_3rd_party_3 |  10.37 M |   4.10 B |

  ### Without indices

  Script size (global logic): 2648 bytes

  | bench                | mem      | cpu      |
  | -------------------- | -------: | -------: |
  | baseline_3rd_party_1 | 197.99 K | 76.81 M  |
  | baseline_3rd_party_2 | 324.01 K | 122.54 M |
  | baseline_3rd_party_3 | 9.79 M   | 3.97 B   |


src/programmable-tokens-onchain-aiken/lib/types.ak
```diff
@@ -23,12 +23,10 @@ pub type ProgrammableLogicGlobalRedeemer {
   ThirdPartyAct {
     /// Index of the registry node in reference inputs
     registry_node_idx: Int,
-    /// List of input indices to process (supports multi-UTxO operations)
-    input_idxs: List<Int>,
     /// Starting index in outputs where processed outputs begin
     outputs_start_idx: Int,
     /// Length of input_idxs list (for validation)
-    length_input_idxs: Int,
+    length_inputs: Int,
```
nemo83: this seems to be now the length of the input list and NOT anymore the length of the index list and the code seems to agree with this. Can you please confirm.

In that case we should change the above comment.

Issue #41, opened 2026-03-20
Programmable tokens can exist outside the mini-ledger


colll78:
See:
https://github.com/cardano-foundation/CIPs/pull/444#issuecomment-4084863264

PR #45, opened 2026-03-31
Fix/minting proof and composability 2

nemo83:
This PR contains both the fix to a vulnerability and an enhancement during minting time.

The issues were raised here: https://github.com/cardano-foundation/CIPs/pull/444#issuecomment-4076182029

And an issue was raised here: https://github.com/cardano-foundation/cip113-programmable-tokens/issues/41

In particular:
1. the `issuance_mint.ak` script it's now "registry token" aware, and requires either the registry node which contains the "programmable token" proof to be either present as ref input (subsequent mint) OR as output (first mint) . Script parameters has been changed to now cater for the registry currency symbol and the redeemer to contain a pointer to the registry ref input OR output.
2. Lifted the need to have during issuance the output of the minted token at first output.


src/programmable-tokens-onchain-aiken/validators/issuance_mint.ak
```diff
@@ -29,34 +33,73 @@ validator issuance_mint(
...
-        fn(token) {
-          let (cs, _tn, _qty) = token
-          cs == own_policy
-        },
-      )
-
-    expect [(_cs, tn, qty)] = tokens_for_policy
+    expect Some(minted_assets) =
+      assets.to_dict(self.mint) |> dict.get(own_policy)
+    expect [Pair(token_name, quantity)] = minted_assets |> dict.to_pairs
```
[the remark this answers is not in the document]
nemo83: This should now be addressed.

src/programmable-tokens-onchain-aiken/validators/issuance_mint.ak
```diff
@@ -29,34 +33,73 @@ validator issuance_mint(
...
+          } else {
+            // Pure subsequent mint, no spend — validate outputs ourselves.
+            validate_mint_outputs(
+              self.outputs,
+              programmable_logic_base,
+              own_policy,
+              token_name,
+              quantity,
+            )?
+          }
```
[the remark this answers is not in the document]
nemo83: Reworked and simplified the logic.

nemo83:
@fabianbormann I've added one last commit to address a bug where the multi asset minting for same policy constraint was not correctly lifted in the registry validator. I've added tests to self document the fix

src/programmable-tokens-onchain-aiken/validators/issuance_cbor_hex_mint.ak
```diff
@@ -16,23 +17,17 @@ validator issuance_cbor_hex_mint(
   mint(_redeemer: Data, own_policy: PolicyId, self: Transaction) {
     trace @"Starting issuance_cbor_hex_mint validation"
     // Must consume the one-shot UTxO
-    let consumed =
+    expect
       list.any(self.inputs, fn(input) { input.output_reference == utxo_ref })
```
fabianbormann, line 21: Wasn't expect not more expensive and you should just pit "list.any(self.inputs, fn(input) { input.output_reference == utxo_ref })" in an and { ... } at the end of the function or is this information outdated logic?

src/programmable-tokens-onchain-aiken/validators/issuance_mint.ak
```diff
@@ -30,38 +35,67 @@ validator issuance_mint(
...
+                input.output.address.payment_credential == programmable_logic_base
+              },
+            )
 
-        // Validate that the output has a stake credential
-        expect Some(Inline(_stake_cred)) = first_output.address.stake_credential
-        and {
-          (first_output.address.payment_credential == programmable_logic_base)?,
-          (quantity_of(first_output.value, own_policy, tn) == qty)?,
+          if plb_spent {
```
fabianbormann: This code is not tested, but I would try to use or over if .. True else .. (if possible here ; if not forget my comment)

or {
    plb_spent,
    validate_mint_outputs(
      self.outputs,
      programmable_logic_base,
      own_policy,
      (self.mint |> assets.tokens(own_policy)),
    )
}

src/programmable-tokens-onchain-aiken/validators/protocol_params_mint.ak
```diff
@@ -22,22 +23,12 @@ validator protocol_params_mint(
   mint(_redeemer: Data, own_policy: PolicyId, self: Transaction) {
     trace @"Starting protocol_params_mint validation"
     // Check 1: This is a one-shot minting policy - must spend the specified UTXO
-    let consumed =
+    expect
       list.any(self.inputs, fn(input) { input.output_reference == utxo_ref })
```
fabianbormann, line 27: same question as above with bringing
list.any(self.inputs, fn(input) { input.output_reference == utxo_ref }) to the end?

src/programmable-tokens-onchain-aiken/validators/registry_spend.ak
```diff
@@ -38,17 +39,12 @@ validator registry_spend(protocol_params_cs: PolicyId) {
...
-        },
-      )
+    expect Some(minted_assets) =
+      assets.to_dict(self.mint) |> dict.get(registry_node_cs)
+    expect [Pair(_, amount)] = dict.to_pairs(minted_assets)
+    expect amount > 0
 
-    // Validation passes if registry NFTs are being minted/burned
-    minting_registry_nodes?
+    True
```
fabianbormann: replace True with amount > 0 instead of the expect above

src/programmable-tokens-onchain-aiken/validators/issuance_cbor_hex_mint.ak
nemo83: fun fact, it's not. 

This is performance over readibility. Otehrwise the `let ... and{}` is ever so slightly less performant but way more readeable

src/programmable-tokens-onchain-aiken/lib/assets.ak
```diff
@@ -154,3 +154,13 @@ fn do_insert(self: Assets, k1: PolicyId, v1: Tokens) -> Assets {
...
+
+pub fn has_policy_nft_strictly(
+  value: Value,
+  policy_id: PolicyId,
+  asset_name: AssetName,
+) -> Bool {
+  expect Some(policy_assets) = assets.to_dict(value) |> dict.get(policy_id)
+  expect [Pair(actual_asset_name, 1)] = dict.to_pairs(policy_assets)
+  actual_asset_name == asset_name
+}
```
KtorZ, line 167: Why not just: https://aiken-lang.github.io/stdlib/cardano/assets.html#has_nft_strict ? 

It does also allow `ada` to be present, but in this scenario, it's not a problem.

src/programmable-tokens-onchain-aiken/validators/issuance_mint.ak
```diff
@@ -30,38 +35,67 @@ validator issuance_mint(
...
+    //   OutputIndex → First mint. PLGlobal doesn't know or see the registry entry for this token.
+    //                 issuance_mint MUST validate outputs. Qty always > 0.
+    //
+    //   RefInput + PLB spent → PLGlobal is mint-aware and will validate
+    //                          everything (mint AND burn). Delegate.
+    //
+    //   RefInput + no PLB spent → Pure subsequent mint. issuance_mint validates.
+    //                             (Burn is impossible here: tokens to burn must
+    //                              be in inputs, prog tokens live at PLB,
+    //                              spending PLB = PLB spent → delegate path.)
```
KtorZ, line 49: How does making this decision part of the redeemer doesn't open up attacks? Couldn't one simply bypass some validations by choosing a different redeemer? 

Sorry, I don't have the full context on the issuance and I am lazily asking you to convince me that this isn't a problem 🙈.

src/programmable-tokens-onchain-aiken/validators/issuance_mint.ak
KtorZ: Agreed, it does read better. 

```ak
or {
    // Can we delegate to PLGlobal? PLGlobal is mint-aware — delegate all validation
    // (handles both minting and burning accounting)
    list.any(
      self.inputs,
      fn(input) {
      },
    )?,
    // No PLB spend → can only be pure minting
    // (burn requires spending PLB UTxOs → plb_spent would be True)
    validate_mint_outputs(
      self.outputs,
      programmable_logic_base,
      own_policy,
      self.mint |> assets.tokens(own_policy),
    )?
}
```

KtorZ: Note: `or` is lazy in its operand, so it only evaluates the second case if the first one is `False`.

src/programmable-tokens-onchain-aiken/validators/issuance_mint.ak
```diff
@@ -118,3 +150,39 @@ fn single_mint_with_credential(
...
+      dict.empty,
+      fn(output, acc) {
+        if output.address.payment_credential == programmable_logic_base {
+          expect Some(Inline(_stake_cred)) = output.address.stake_credential
+          tokens.union(acc, assets.tokens(output.value, own_policy))
+        } else {
+          acc
+        }
+      },
+    )
```
KtorZ, line 170: This is: https://github.com/cardano-foundation/cip113-programmable-tokens/blob/main/src/programmable-tokens-onchain-aiken/validators/programmable_logic/transfer.ak#L90-L106

We could factor this out under `lib/assets` and reuse it.

src/programmable-tokens-onchain-aiken/lib/assets.ak
nemo83: The problem is "other assets are not tolerated". This function is used to check `mint` value, where multiple nfts could be minted, so the strictly is _too much_, at the same time the non _strictly_ is too little as we have an additional check from this check where only one asset for the given policy can be present. 

The scenario is the "cip113 bootstrap transaction" where multiple nfts are minted.

src/programmable-tokens-onchain-aiken/validators/issuance_mint.ak
nemo83: The redeemer is used to point to an output or a ref input to save execution costs (the script could find one or the other by itself but it would be very expensive). Then depending on where is the "registry node" and whether there are also programmable logic being spent, the invariances are verified by either the spending script (which is mint aware) or the issuance script (which is NOT transfer aware).

In a very special corner case, ie. Transfer A, B, C but mint (for the very first time) D, both scripts run full logic.

src/programmable-tokens-onchain-aiken/lib/assets.ak
KtorZ: Ah, I see. You want to allow other policies.

PR #50, opened 2026-04-22
chore: added publish handler to allow registration with conway cert

nemo83:
In order to allow script 's stake address to be registered with conway-like certs, the publish endpoint has been added.

I can't see any potential vulnerability.


validators/programmable_logic_global.ak
```diff
@@ -49,6 +50,10 @@ validator programmable_logic_global(params_policy: PolicyId) {
     }
   }
 
+  publish(_r: Data, _c: Certificate, _t: Transaction) {
+    True
```
KtorZ: That'd allow anyone to *unregister* the credential as well (and not only claim back the deposit, but also cause a denial of service).

At the very least, we should limit it to registering:

```aiken
   expect RegisterCredential{..} = c
```

Now, that prevents the credentials from ever being unregistered. Which is _arguably bad_ because that's just polluting the chain in the long run. So we could restrict unregistering and delegating to being only allowed by the 3rd party credential?

nemo83: That's a great catch! 

There should always ever gonna be one of these scripts, so fail on unregister should be okay.

KtorZ (approved):

PR #51, opened 2026-05-13
Fix audit findings 3, 8 and 9

### LINKED ISSUE #55: Registry init does not bind the origin node to `registry_spend`
# Finding 03: Registry init does not bind the origin node to `registry_spend`

## Description

On `RegistryInit`, `registry_mint` calls `validate_directory_init`, which checks the one-shot spend, that no registry inputs are spent, a single node output with correct value/datum/NFT shape, origin list semantics, and the origin registry NFT in the mint.  
It does **not** require that output to be locked at the **`registry_spend`** address. A mistaken bootstrap tx can therefore place the origin registry node at any address while still passing validation.

## Relevant code snippet

In particular, `collect_node_ios` anchors node outputs to the address of the first node input, but only if such an input already exists:

```aiken
when node_inputs is {
  [first, ..] -> {
    let expected_address = first.output.address
    expect
      list.all(
        node_outputs,
        fn(output) { output.address == expected_address },  // check that node outputs are at the same address as the first node input
      )
    (node_inputs, node_outputs)
  }
  [] -> ([], node_outputs)
}
```

This means `RegistryInsert` preserves the address of the covering node input, but `RegistryInit` has no such anchor because there are no existing node inputs yet.

## Exploit scenarios

- During bootstrapping (`RegistryInit`), the operator sends the origin registry node to an address they control instead of the canonical `registry_spend` address.
- Later, they can perform `RegistryInsert`. In `RegistryInsert`, the validator only enforces that node outputs stay at the same address as the first node input, so the transaction validation pass with newly inserted registry nodes also remain at the attacker-controlled address.
- They can then send an origin node UTxO back to the `registry_spend` address with the correct NFT and output format.

At that point, some authentic registry NFTs have existed outside the canonical registry lock.
If the protocol continues operating and more nodes are registered or more programmable tokens are minted, this opens two exploit scenarios.

### Scenario 1: Authority hijack via poisoned registry node

- The attacker uses the registry NFTs they control to create or reintroduce a registry node with a malicious datum.
- In that forged datum, they can replace `third_party_transfer_logic_script` with a script they control.
- PLG will trust that registry node as the authority for that policy.
- The attacker can then invoke `ThirdPartyAct` and seize users' programmable-token UTxOs for that policy, subject to the remaining accounting checks.

### Scenario 2: Programmable-enforcement bypass via forged non-membership interval

- The attacker forges registry node keys / gaps so that a genuinely registered policy can be treated as if it were not included in the registry.
- This allows them to construct a fake `TokenDoesNotExist`-style non-membership proof.
- PLG then treats that policy as non-programmable and skips the normal programmable-token containment checks for it.
- As a result, the token can be moved outside programmable custody in otherwise authorized transfer flows.

## Impact

This is not a permissionless outsider issue. Exploitation requires malicious or compromised control of the bootstrap flow, or collusion with the party that receives the misdirected origin node.

However, if that happens, the impact is high:

- The registry custody invariant is broken at genesis
- Authentic registry NFTs can remain under attacker-controlled custody outside the canonical `registry_spend` lock
- Poisoned registry nodes can later be reintroduced and trusted by downstream validators, and
- The attacker can abuse that trust either to hijack third-party authority for a policy or to make a registered policy appear absent from the registry.

This is therefore a bootstrap / insider trust failure with protocol-wide consequences: once the canonical registry lineage is broken, downstream validators may continue trusting poisoned registry state.

## Recommendation

Enforce the registry custody address on-chain during `RegistryInit`.

Specifically:

- Store the `registry_spend` address or payment credential in a protocol param UTxO datum
- Include the protocol-params currency symbol as a parameter of `registry_mint`,
- Require a protocol-params reference input in `RegistryInit`,
- Read the expected `registry_spend` address or payment credential from that protocol-params datum, and
- Assert that the origin registry node output is locked at that expected address.

This ensures the registry lineage starts under the canonical registry lock and prevents attacker-controlled or non-canonical custody from being established at bootstrap.

### LINKED ISSUE #60: Inefficient Membership Check in `issuance_mint` Validator
# Finding 8: Inefficient Membership Check in `issuance_mint` Validator

## Description

In the `issuance_mint` validator, there is a check for whether the `minting_logic_cred` is invoked during registration. This check is currently implemented by traversing the list of withdrawals in the transaction and creating an intermediate list of credentials, which is then checked for membership against the `minting_logic_cred`. This approach is inefficient as it traverses the list of withdrawals twice and allocates an intermediate list, which can lead to increased execution cost.

## Relevant Code Snippet

```rs
// Collect all invoked scripts (withdraw-zero pattern)
let invoked_scripts =
  list.map(
    self.withdrawals,
    fn(wdrl) {
      let Pair(cred, _amount) = wdrl
      cred
    },
  )
```

Later in the code, there is a check for membership:

```rs
// substandard minting invoked?
list.has(invoked_scripts, minting_logic_cred)?,
```

## Recommendation

Instead of creating an intermediate list of credentials, we can directly check for membership in the original list of withdrawals without creating a new one. This can be done by iterating through the list of withdrawals and checking if any of the credentials match the `minting_logic_cred` directly, which would be more efficient and reduce execution cost.

Implementation of the recommended change could look like this:

```rs
let minting_logic_invoked =
  list.any(
    self.withdrawals,
    fn(wdrl) {
      let Pair(cred, _amount) = wdrl
      cred == minting_logic_cred
    },
  )
```

Then, instead of checking `list.has(invoked_scripts, minting_logic_cred)`, we can directly check the `minting_logic_invoked` variable:

```rs
// substandard minting invoked?
minting_logic_invoked,
```

This change eliminates the need for an intermediate list and reduces the number of traversals, improving the efficiency of the validator.

---

### LINKED ISSUE #61: Indirect Delegation Signal in PLB-Input Scan
# Finding 9: Indirect Delegation Signal in PLB-Input Scan

## Description

In the `issuance_mint` validator, there is a condition that checks for the presence of a PLB input in the transaction's inputs to determine if mint custody is being delegated to PLGlobal. This check is implemented by scanning through the transaction inputs and checking if any of the input addresses have a payment credential that matches the `programmable_logic_base`.

## Relevant Code Snippet

```rs
// Can we delegate to PLGlobal?
// PLGlobal is mint-aware — delegate all validation
// (handles both minting and burning accounting)
list.any(
  self.inputs,
  fn(input) {
  },
),
```

## Recommendation

Instead of scanning through all transaction inputs to check for a PLB input, it is more efficient to check for the presence of the PLGlobal delegation credential directly in the transaction's withdrawals. This provides a more direct signal of delegation to PLGlobal and reduces execution cost.

However, checking for the PLGlobal delegation credential in the withdrawals instead of the PLB input may allow for a scenario where the PLGlobal delegation credential is included in a transaction without including a PLB input. Therefore, it is important to ensure that no critical logic in the PLGlobal can be bypassed by not including a PLB input.

**Note:** Special attention should be given to the PLGlobal `ThirdPartyAct` path, which has some documented issues.

## References

[Finding 04: Escape of Minted Tokens during Third-Party Action](./04-third-party-act-can-falsely-satisfy-issuance-delegation.md)

---

nemo83:
Fixes #55 
Fixes #60 
Fixes #61 

Addresses three numbered audit findings — audit/03 (registry init custody binding), audit/08 (inefficient
membership check), audit/09 (indirect delegation signal) — plus a slate of registry-side hardening surfaced while
reviewing the recommendations. Stdlib also bumped from 3.0.0 to 3.1.0 (d3279b9) along the way; the dict.expect_tail
 helper introduced there is used in the tightened validate_directory_node_output.

Audit findings addressed

audit/03 — Registry init does not bind origin node to registry_spend — Resolved

registry_mint is now parameterised on registry_spend_cred: Credential. validate_directory_init
(lib/linked_list.ak:124) asserts the origin output's payment credential matches it, and the RegistryInsert branch
(validators/registry_mint.ak:92-93) independently asserts the covering input is at the same address — closing the
"Scenario 1/2" cascade described in the finding even if the inductive chain (origin → covering → inserts) is broken
 by a future alternative spend script.

Implementation note. The finding recommended threading the credential through a protocol-params UTxO datum +
reference input. We instead promoted it to a direct script parameter. On-chain invariant is identical; both
approaches share the same off-chain trust assumption (deployer must use the genuine registry_spend hash). Parameter
 approach is one fewer reference input at Init and bakes the binding into the script hash.

audit/08 — Inefficient membership check in issuance_mint — Resolved

The intermediate invoked_scripts list is gone. Withdrawals are scanned exactly once via list.foldl2
(validators/issuance_mint.ak:33-45), detecting both minting_logic_cred and plg_stake_cred membership in the same
pass. Strictly better than the recommended list.any because it folds in the second check the Finding 09 fix needs.

audit/09 — Indirect delegation signal — Resolved (depends on audit/04)

issuance_mint takes a new plg_stake_cred: Credential parameter. The PLB-input scan is replaced by a direct check
that PLGlobal appears in tx.withdrawals (plg_invoked, line 92).

Residual dependency. Finding 09's own text warns: "it is important to ensure that no critical logic in the PLGlobal
 can be bypassed by not including a PLB input. Special attention should be given to the PLGlobal ThirdPartyAct
path, which has some documented issues." 

Additional hardening (not tracked under numbered findings)

Surfaced while reviewing audit/03 and during the registry-question follow-up; commit messages on the branch have
the full per-item rationale.

- Strict registry-NFT identity in verify_registry_node (0f18345). Replaced assets.has_any_nft with strict
assets.has_nft(value, registry_node_cs, own_policy) — binds the NFT asset name to the datum key directly instead of
 trusting the linked-list invariant transitively.
- Named sentinels for the linked list (2e523b0). origin_node_key = #"" and sentinel_next_key (30-byte all-ff,
intentionally longer than a 28-byte policy id, doc-commented as load-bearing) in lib/registry_node.ak. Replaces
inline literals in validate_directory_init.
- Stricter validate_directory_node_output (2e523b0, refined further by stdlib 3.1.0 migration in d3279b9). Eight
additional invariants, including two CIP-113 §Registration MUSTs that were not previously checked: address must
have only a payment credential (point 7) and no reference script (point 6). Tightened the loose tn == node.key ||
tn == origin_node_tn disjunct to strict equality (only attacker-visible behaviour change in the function; origin
still passes via #"" == #"").
- Defense-in-depth address check in RegistryInsert (2e523b0). Restates the registry-custody invariant at the Insert
 site so the path doesn't rely on the origin → covering → inserts chain being unbroken by future contracts.


validators/registry_mint.ak
```diff
@@ -83,28 +86,39 @@ validator registry_mint(
...
         // Must mint exactly one node token with the new key
-        let just_single_mint = assets.has_nft(self.mint, policy_id, key)
+        expect validate_mint(self.mint, policy_id) == key
 
-        // Parse the outputs to find the two new nodes
-        let output_nodes =
-          list.map(
+        // Must have exactly 2 outputs: the updated covering node and the new inserted node
+        expect list.length(node_outputs) == 2
+        // Single-pass scan of node_outputs: validate each output's common
```
KtorZ: Technically that's already two pass since there's a `list.length` just above 😬. If we know the exact count of node outputs (i.e. 2), why not just:

```ak
 expect [node_output_1, node_output_2] = node_outputs
```

and then simply:

```ak
let node1 = validate_directory_node_output(node_output_1, policy_id)
let node2 = validate_directory_node_output(node_output_2, policy_id)

let updated = if is_inserted_directory_node(node1, key, covering_node.next) {
  node2
} else {
  expect is_inserted_directory_node(node2, key, covering_node.next)
  node1
}

expect is_updated_directory_node(updated, covering_node, key)
```

KtorZ (commented):
Are you able to add test cases that illustrates the solved problems?

lib/linked_list.ak
```diff
@@ -5,40 +5,52 @@ use cardano/address.{Credential, Script, VerificationKey}
...
-  // Must have exactly 2 unique tokens (Ada + node token)
-  expect count_unique_tokens(value) == 2
-
-  // Must have exactly 1 node token from this policy
-  let node_tokens = tokens(value, node_cs)
-  expect dict.size(node_tokens) == 1
+  // Address must have only a payment credential (CIP-113 §Registration point 7)
+  expect output.address.stake_credential == None
+  // No reference script (CIP-113 §Registration point 6)
+  expect output.reference_script == None
```
KtorZ:
```suggestion
  expect None = output.address.stake_credential
  // No reference script (CIP-113 §Registration point 6)
  expect None = output.reference_script
```

lib/linked_list.ak
```diff
@@ -111,46 +137,45 @@ pub fn is_inserted_directory_node(
...
+/// Strict 28-byte credential check. Used by `is_inserted_directory_node` to
+/// reject the length-0 origin shape on a non-origin node.
+pub fn is_28_byte_credential(cred: Credential) -> Bool {
   when cred is {
-    VerificationKey(hash) -> is_valid_credential_length(hash)
-    Script(hash) -> is_valid_credential_length(hash)
+    VerificationKey(hash) -> bytearray.length(hash) == 28
+    Script(hash) -> bytearray.length(hash) == 28
   }
 }
```
KtorZ: Given all the `bytearray.length(_) == 28` in this snippet, it might be worth moving to a dedicated function `is_28_bytes` that does both so that the compiler can more easily hoist that function instead of duplicating builtin calls.

lib/linked_list.ak
nemo83: why the swap? expect actual is eq to expected 

or is it efficiency?

validators/registry_mint.ak
nemo83: yeah, makes sense.

one a side note, would it make sense to add 
* list.expect_none
* list.expect_all
?

lib/linked_list.ak
KtorZ: They're both equivalent, but the latter is better code-gen wise; In one case it's an equality between two objects, in the other it's a pattern-match.

validators/registry_mint.ak
KtorZ: Not really, since in both cases, you do have to traverse the entire list anyway. So you can't really short-circuit the function in the happy path; only when the element is not found. The "worse case" scenario is the happy path where you have to look at every element.

nemo83: Refactored code to implement true one inputs/outputs traversals. Removed `linked_list.collect_node_ios`, and implemented specific assertions in `RegistryInit` and `RegistryInsert` branches.

Added tests to ensure finding is now fixed and no regressions are introduced.

lib/linked_list.ak
nemo83: Done ✅

nemo83:
Updated to stdlib 3.1.0 as I could immediately take advantage of the new fail fast `list.expect` features.

Will update to latest aiken version in a separate PR.

KtorZ (approved):
Looks sensible, though I haven't spend all the time I would've liked on the review; but I lack of it.

PR #70, opened 2026-05-26
refactor(registry): Drop Redundant Length and Ordering Checks in RegistryInsert (Finding 19)

### LINKED ISSUE #74: Redundant Checks and Misleading Variable Names in RegistryInsert Branch
# Finding 19: Redundant Checks and Misleading Variable Names in RegistryInsert Branch

In the `RegistryInsert` branch of the `registry_mint` validator, there are several checks that are redundant under the current assumptions and helper definitions. These redundancies can be removed to simplify the code without affecting its correctness. Below are the identified redundancies and the rationale for their removal:

## 1. The singleton pattern:

In the `validate_directory_node_output` function, the check `expect dict.size(node_tokens) == 1` is redundant because the line `expect [Pair(tn, qty)] = dict.to_pairs(node_tokens)` already proves that `node_tokens` has exactly one token name. Therefore, the size check can be removed.

```diff
  let node_tokens = tokens(value, node_cs)
- expect dict.size(node_tokens) == 1

  expect node: RegistryNode = datum
  expect bytearray_lt(node.key, node.next)
- let token_pairs = dict.to_pairs(node_tokens)
- expect [Pair(tn, qty)] = token_pairs
+ expect [Pair(tn, qty)] = dict.to_pairs(node_tokens)
  expect qty == 1
```

## 2. The exception for the origin node:

In the same function `validate_directory_node_output`, the check `expect tn == node.key || tn == origin_node_tn` is redundant because `origin_node_tn` is defined as `#""`, and the `init` function already requires that the origin node’s `node.key == #""`. Therefore, we can simplify this check to just compare `tn` with `node.key`.

```diff
- expect tn == node.key || tn == origin_node_tn
+ expect tn == node.key
```

Additionally, in the `RegistryInit` branch, in the `validate_directory_init` function, the check `expect node.key == #""` can be replaced by `expect node.key == origin_node_tn` for better readability and maintainability, as it uses the defined constant for the origin node token name (defined in `lib/registry_node.ak`).

```diff
- expect node.key == #""
+ expect node.key == origin_node_tn
```

This change secures the code against future modifications where the origin node token name might be changed, ensuring that all references to it are consistent.

## 3. The length check for the key:

In the main `RegistryInsert` branch, the check `(bytearray.length(key) == 28)?` is redundant because the `is_token_registered?` helper already requires that the computed policy ID matches the key, and the `apply_hashed_parameter` function returns a 28-byte policy ID. Therefore, if the computed policy ID matches the key, the key must already be 28 bytes long.

```diff
  and {
    is_token_registered?,
-   (bytearray.length(key) == 28)?,
    ...
  }
```

## 4. The ordering checks for the covering node:

In the same main `RegistryInsert` branch, the checks `bytearray_lt(covering_node.key, key)?` and `bytearray_lt(key, covering_node.next)?` are redundant. Since the `validate_directory_node_output` function already checks that `node.key < node.next`, and the helper definitions for `registry_node_updated?` and `registry_node_inserted?` enforce the relationships between the covering node and the new key, these checks can be removed.

```diff
  and {
    is_token_registered?,
-   (bytearray.length(key) == 28)?,
-   bytearray_lt(covering_node.key, key)?,
-   bytearray_lt(key, covering_node.next)?,
    just_single_mint?,
    (list.length(output_nodes) == 2)?,
    registry_node_updated?,
    registry_node_inserted?,
  }
```

## 5. The `just_single_mint?` check:

In the `RegistryInsert` branch, the `just_single_mint?` check is technically redundant under the following stronger assumptions:

```diff
expect [covering_input] = node_inputs
expect [Pair(tn, qty)] = dict.to_pairs(node_tokens)
expect qty == 1
expect tn == node.key
list.length(output_nodes) == 2
registry_node_inserted?
registry_node_updated?
```

The above checks imply that the inserted output must carry one registry NFT whose token name is `key`. Since the covering input did not have token name `key`, the ledger balance requires the transaction to mint that NFT. However, since `just_single_mint?` is cheap, clear, and provides a direct check for the minting of the node token with the new key, it may be worth keeping for clarity and local reasoning about the minting behavior.

## 6. Computational redundancy in output node validation:

Currently, the code builds all output nodes, scans them twice with `list.any`, and separately checks their length. This can be refactored to validate the output nodes in a single pass while checking their order.

Replace the following code:

```diff
-        // Parse the outputs to find the two new nodes
-        let output_nodes =
-          list.map(
-            node_outputs,
-            fn(out) { validate_directory_node_output(out, policy_id) },
-          )
-
-        // One output must be the updated covering node (key -> insert_key)
-        let registry_node_updated =
-          list.any(
-            output_nodes,
-            fn(node) { is_updated_directory_node(node, covering_node, key) },
-          )
-
-        // One output must be the new inserted node (insert_key -> next)
-        let registry_node_inserted =
-          list.any(
-            output_nodes,
-            fn(node) {
-              is_inserted_directory_node(node, key, covering_node.next)
-            },
-          )
+        // Exactly two registry node outputs are expected:
+        // one updated covering node and one newly inserted node.
+        expect [node_output_a, node_output_b] = node_outputs
+
+        let node_a =
+          validate_directory_node_output(node_output_a, policy_id)
+
+        let node_b =
+          validate_directory_node_output(node_output_b, policy_id)
+
+        // Outputs may appear in either order. (Alternatively, we could require a specific order and simplify this check.)
+        let registry_nodes_valid =
+          if node_a.key == covering_node.key {
+            and {
+              is_updated_directory_node(node_a, covering_node, key)?,
+              is_inserted_directory_node(node_b, key, covering_node.next)?,
+            }
+          } else {
+            and {
+              is_inserted_directory_node(node_a, key, covering_node.next)?,
+              is_updated_directory_node(node_b, covering_node, key)?,
+            }
+          }
...
-          // Must have exactly 2 outputs: the updated covering node and the new inserted node
-          (list.length(output_nodes) == 2)?,
-          registry_node_updated?,
-          registry_node_inserted?,
+          registry_nodes_valid?,
```

This refactor removes the separate length check and the two `list.any` traversals while preserving the behavior that the outputs may appear in either order. It also makes the validation logic more direct and easier to follow.

## 7. Misleading naming of variables:

In the `RegistryInsert` minting path there are several names that are misleading enough to make review harder, or cause confusion about the code’s intent:

- `is_token_registered`: This name suggests that the function checks if a token is registered in the registry, but it actually checks if the key is a valid official programmable-token registration candidate by verifying that the computed policy ID matches the key and that the policy is minting in this transaction. A more accurate name could be `is_valid_registration`, `valid_registration`, or `is_official_issuance_policy_mint`.
- `just_single_mint`: This name suggests that the entire minting action involves exactly one token, but it only checks that the registry NFT for the key is minted under the registry policy. A clearer name could be `inserted_node_nft_minted`, `new_registry_nft_minted`, or `has_insert_key_nft_mint`.

nemo83:
Fix: #74 

## Summary

Cleans up Audit Finding 19 (informational, code quality). Three checks at the tail of the `RegistryInsert` branch are redundant — transitively implied by upstream checks — and
  are removed.

  ## What's removed

  - `(bytearray.length(key) == 28)?` — implied by `is_programmable_token_id_valid`, which equates `key` to the output of `apply_hashed_parameter` (always `blake2b_224` → 28
  bytes).
  - `bytearray_lt(covering_node.key, key)?` — implied by `validate_directory_node_output`'s per-output `node.key < node.next` check, given that `is_updated_directory_node` pins
  `updated.key == covering_node.key` and `updated.next == key`.
  - `bytearray_lt(key, covering_node.next)?` — same shape on the inserted output: `inserted.key == key`, `inserted.next == covering_node.next`.

  Two new `fail` tests pin the upstream invariants as regression guards — they construct scenarios where the *only* violation is the targeted inequality and verify the validator
  still rejects via `validate_directory_node_output`. If that upstream check is ever weakened, the alarm fires.


validators/registry_mint.ak
```diff
@@ -177,13 +176,23 @@ validator registry_mint(
...
+        //   - the two ordering checks are implied by
+        //     `validate_directory_node_output`'s `node.key < node.next`
+        //     check, combined with the relationships `updated.key ==
+        //     covering_node.key`, `updated.next == key`, `inserted.key ==
+        //     key`, `inserted.next == covering_node.next` enforced by
+        //     `is_updated_directory_node` and `is_inserted_directory_node`.
+        // See registry_mint.test.ak `registry_insert_fails_covering_key_ge_*`
+        // / `_insert_key_ge_covering_next` for explicit upstream-coverage
+        // tests.
+        True
```
fabianbormann: Maybe I'm wrong, but wouldn't it make sense to replace

```aiken
expect is_updated_directory_node(updated, covering_node, key)
True
```

simply by

```aiken
is_updated_directory_node(updated, covering_node, key)
```

as is_updated_directory_node already returns a bool

fabianbormann (changes requested):

fabianbormann (approved):

PR #77, opened 2026-05-28
fix(transfer): Finding 02 — pure mints no longer subject to transfer-logic enforcement

### LINKED ISSUE #54: Enforcement of `transfer_logic_script` for Purely Minted Tokens
# Finding 02: Enforcement of `transfer_logic_script` for Purely Minted Tokens

## Description

The current design and implementation of CIP-113 contain a contradiction regarding when the `transfer_logic_script` is enforced. The intended design is that the `transfer_logic_script` should be enforced whenever a token is transferred (i.e., whenever the token is in the input from a smart-wallet) and should not be enforced when the token is purely minted (i.e., the token is only in the mint field and not in the input from a smart-wallet).

However, the current implementation contains a contradiction:

1. In the `validators/issuance_mint.ak`, the `transfer_logic_script` is not enforced when minting a token, which is correct and aligns with the intended design.
2. However, in the `validators/programmable_logic_global.ak` (specifically in the `validate_transfer` function), the `transfer_logic_script` is enforced if a token is purely minted, which is incorrect and contradicts the intended design.

Relevant code snippet from `validators/programmable_logic_global.ak`:

```ak
/// Validate a transfer action
pub fn validate_transfer(
  self: Transaction,
  prog_logic_cred: Credential,
  get_registry_node: fn(Int) -> Data,
  proofs: List<RegistryProof>,
) -> Bool {
  ...
  let input_assets = collect_input_assets(self, has_withdrawal, prog_logic_cred)
  ...
}
```

In this code, the `collect_input_assets` function collects the tokens from the transaction inputs that are sent from a smart-wallet (programmable + non-programmable), but it also collects the tokens that are being minted in the same transaction (programmable + non-programmable). This means that if a token is being purely minted (i.e., it is only in the mint field and not in the input from a smart-wallet), it will be subject to the `transfer_logic_script` even though it is not being transferred, which is a contradiction to the intended design.

```ak
/// Get inputs from prog_logic_cred and verify ALL have proper authorization
/// This function enforces ownership by requiring that every input from prog_logic_cred
/// is authorized by its stake credential (signature or script invocation).
/// If ANY input lacks authorization, the transaction FAILS.
/// Sum values from a list of inputs
/// Check transfer logic for each currency symbol and filter to only programmable tokens
fn verify_proofs(..) -> List<RegistryProof> {..}
```

`verify_proofs` is mainly doing two things:

1. Enforcing `transfer_logic_script` execution for all programmable tokens in the transaction, including purely minted tokens.
2. Token balance check by comparing the tokens in the inputs and outputs (inputs must be in the outputs, i.e., the programmable tokens never leave the smart-wallet).

   ```ak
   expect
     tokens.contains(
       dict.to_pairs(head_output_tokens),
       dict.to_pairs(head_input_tokens),
     )
   ```

Moreover, requiring a proof for purely minted tokens creates redundancy and inefficiency, since the minting policy is already performing checks to ensure that the token is being minted correctly.

## Impact

1. **Design Contradiction**: The enforcement of the `transfer_logic_script` for purely minted tokens contradicts the intended design of CIP-113.
2. **Unintended Consequences**: This could lead to unexpected behavior or errors in transactions involving purely minted tokens, as the `transfer_logic_script` may impose constraints or requirements that are not relevant to the minting process. The `transfer_logic_script` may require certain conditions to be met (e.g., specific signatures, state checks) that may not be desirable or necessary for purely minted tokens, leading to failed transactions or unintended consequences for users minting new tokens.
3. **Inefficiency**: Requiring the `transfer_logic_script` to be enforced for purely minted tokens creates unnecessary overhead and inefficiency, as it requires additional script execution for tokens that are not being transferred. Moreover, it requires the user to provide proofs for purely minted tokens, which is redundant since the minting policy already performs necessary checks during the minting process.

## Recommendations

In order to resolve this contradiction, the implementation should be updated to ensure that the `transfer_logic_script` is only enforced for tokens that are being transferred (i.e., tokens that are in the input from a smart-wallet) and not for tokens that are purely minted (i.e., tokens that are only in the mint field and not in the input from a smart-wallet). This could involve updating the `validate_transfer` function to differentiate between transferred tokens and purely minted tokens, ensuring that the `transfer_logic_script` is only applied to the former.

To achieve this, the following changes could be made:

1. Separate the 2 main functionalities in `verify_proofs` into two functions: one for enforcing `transfer_logic_script` execution and another for token balance checks. This would allow for more granular control over when the `transfer_logic_script` is enforced.
2. In the function responsible for enforcing `transfer_logic_script` execution, ensure that it only processes tokens that are in the inputs from a smart-wallet.
3. In the function responsible for token balance checks, ensure that it accounts for purely minted tokens appropriately, without subjecting them to the `transfer_logic_script`.

By implementing these changes, the design would align with the intended functionality of CIP-113, ensuring that the `transfer_logic_script` is enforced only for transferred tokens and not for purely minted tokens, thus resolving the contradiction in the current implementation. It would also improve the efficiency of the implementation by avoiding unnecessary enforcement of the `transfer_logic_script` and proof requirements for purely minted tokens.

---

nemo83:
Fixes #54 

## Summary

Closes Finding 02 from the audit (`audit/02-transfer-logic-enforcement-contradiction.md`). `programmable_logic_global.validate_transfer` previously folded `tx.mint` into the same accumulator as PLB inputs, so the per-policy `transfer_logic_script` enforcement and `TransferAct` proof requirement applied to every policy in `tx.mint` — including policies that only appear in mint and not in any PLB input. `issuance_mint` had already authorised those pure mints, so the enforcement was redundant.

After this PR, pure-mint policies are dropped from `input_assets` before `verify_proofs` runs — they are `issuance_mint`'s responsibility, not the transfer path's.

## The change

**Three small, focused changes (5 commits):**

1. **`lib/assets.ak`** — adds `assets_split_at`, a thin public wrapper over the existing `do_split_at` so callers can split an `Assets` directly (the existing public `split_at` only takes a `Value`).
2. **`validators/programmable_logic/transfer.ak`** — `collect_input_assets` now seeds with `[]` (was `assets.from_value(tx.mint)`); a new private helper `apply_mint_to_known_policies` folds mint deltas in afterward, but only for policies that appear in PLB inputs. Mint-only policies are dropped. `verify_proofs` is unchanged.

   The helper uses `assets.assets_split_at` for the merge — same `split_at`/`pop_until` idiom already used in `verify_proofs` and `third_party.ak`, so the transfer path now has one consistent "find-policy-in-sorted-Assets" pattern.
3. **`validators/programmable_logic_global.test.ak`** — three pre-existing tests retitled/un-`fail`ed because their assertions implicitly pinned the buggy behaviour, plus one new `fail` test pinning the strict 1-proof-per-input-policy contract (so providing a surplus
proof for a pure-mint policy is rejected).

**Off-chain consequence:** a pure-mint policy must NOT carry a `TransferAct` proof. The strict 1-proof-per-input-policy contract is preserved; surplus proofs become leftover and are rejected by the final `[] == verify_proofs(...)` check.

## Cost

`aiken bench -m programmable_logic/benchmarks`, sizes 0..30 — deltas vs `main`:

| benchmark | mem @ 30 Δ | cpu @ 30 Δ |
|---|---:|---:|
| `many_tokens` | +0.52% | +0.34% |
| `many_inputs` | +0.13% | +0.08% |
| `many_policies` | +0.09% | +0.06% |
| `many_outputs` | +0.15% | +0.10% |

- **Constant**: small one-time overhead from the new helper call (~2K mem, ~470K cpu) even when `tx.mint` is empty.
- **Per-step on every axis: 0.** The dominant ~23.5M cpu / unique policy slope is unchanged.
- Pure-mint-heavy transactions are strictly *cheaper* after this PR (those policies no longer enter `verify_proofs`).


fabianbormann (approved):

validators/programmable_logic/transfer.ak
```diff
@@ -65,27 +69,105 @@ fn collect_input_assets(
...
+          let
+            _before,
+            mint_at_policy,
+            mint_after,
+          <- assets.assets_split_at(mint, i_policy)
+          let merged =
+            if mint_at_policy == dict.empty {
+              i_tokens
+            } else {
+              tokens.union(i_tokens, mint_at_policy)
```
KtorZ:
```suggestion
              tokens.union(mint_at_policy, i_tokens)
```

Minted tokens will generally be smaller than `i_tokens` I believe; so that is likely slightly better (left side is consumed into the right side).

KtorZ (approved):

nemo83:
neat changes, thanks @KtorZ

PR #78, opened 2026-06-05
 feat(unfracking): Finding 17 — Unfracking action for same-owner PLB UTxO restructuring

### LINKED ISSUE #72: Unconstrained Asset Addition in Smart-Wallets
# Finding 17: Unconstrained Asset Addition in Smart-Wallets

## Description

In the `ThirdPartyAct` path of the `programmable_logic_global` validator, the `validate_3rd_party` function checks that the target policy in the value of the continuing outputs is changed, but does not impose any constraints on the number of assets that can be added.

If a large number of assets are added to the continuing outputs, it may lead to an increase in execution costs when consuming that UTxO in the future. However, this increase is not problematic for the current implementations of the `ThirdPartyAct` path, as the `validate_3rd_party` function does not traverse the assets with the same policy.

However, this issue could become problematic if the execution cost of the `third_party_transfer_logic_script` for the target policy depends on the number of assets with the same policy. Or the `transfer_logic_script` execution cost depends on the number of assets in the value of the continuing outputs.

This problematic state can be reached not only by the `ThirdPartyAct` path, but also by any other path that allows adding assets to the continuing outputs without constraints. For example, during minting, transfer, or both of them, a large number of assets can be added to the continuing outputs, which can lead to the same issue.

## Impact

A UTxO with a large number of assets can lead to increased execution costs when consuming that UTxO in the future, which can be problematic for users and applications interacting with the system.

The value in a single UTxO can have approximately 1230 assets (a value with 1 ADA, 1 token with policy A, and 1230 tokens with policy B).

## Benchmarks

```plaintext
    ┍━ assets.test ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
    │ PASS [mem:  33.47 K, cpu:  12.82 M] split_at_test_attacker_add_1230_tokens_same_policy
    │ PASS [mem:  39.23 K, cpu: 349.69 M] split_at_test_victim_remove_target_token_different_policy
    ┕━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 2 tests | 2 passed | 0 failed
```

In this benchmark, attacker adds 1230 tokens (Policy B), then victim removes the target token (Policy A). The execution cost of the victim's transaction is significantly higher than the attacker's transaction, which demonstrates the potential increase in execution costs when consuming a UTxO with a large number of assets.

These numbers only represent the execution cost of the `split_at` function, which is only a part of the `ThirdPartyAct` path. The total execution cost of the `ThirdPartyAct` path can be significantly higher, especially if the `third_party_transfer_logic_script` execution cost increases with such a state.

## Recommendation

1. To mitigate this issue, we recommend introducing an `Unfracking` action in the PLGlobal redeemer. This action would allow users to redistribute the tokens over multiple UTxOs without fundamentally changing the distribution of value and ownership. This action should be separate from the `ThirdPartyAct` path, as it should not be used for compliance operations, but rather as a tool for users to manage their UTxOs and avoid fragmentation issues.

2. Alternatively, if the `Unfracking` action is not introduced, we recommend imposing constraints on the number of assets that can be added to the continuing outputs in any path that allows adding assets. This could involve setting a maximum limit on the number of assets or implementing a cost mechanism that increases with the number of assets added. Such paths include the `ThirdPartyAct` path, minting paths, and transfer paths. Note, setting the maximum limit to a value that is too low may limit the usability of the smart-wallets, while setting it to a value that is too high may not effectively mitigate the issue. Therefore, careful consideration should be given to determining an appropriate limit.

nemo83:
Fixes #72 

## Summary

Closes audit Finding 17 (`audit/17-unconstrained-asset-addition-in-smart-wallets.md`) by adding an `Unfracking` redeemer to `programmable_logic_global` that lets a holder restructure their own PLB UTxOs without invoking any substandard transfer logic. The validator's contract is minimal — same owner across all PLB inputs and outputs, per-policy totals conserved at PLB, no mint — so the holder is free to pick any output layout they want.

The typical use case (and the audit-finding motivation) is breaking up a multi-policy "fracked" UTxO into single-policy outputs, so that a freeze (or any per-policy compliance action) stops co-locating its blast radius onto unrelated policies in the same UTxO. But the validator itself accepts any layout — single-policy, partial split, identity reshape, multi-input consolidation, anything the holder needs. Where the owner is a script, it's still required for the contract to execute.

Unfracking is **same-owner-only**: all PLB inputs and PLB outputs must share one stake credential, pinned by the first PLB input encountered. Cross-owner moves are transfers, by definition, and require their substandard's `transfer_logic_script` — they belong to `TransferAct`, not here.

## What Unfracking permits / forbids

**Permits:**
- Spending one or more PLB inputs the caller is authorised over — vkey signatory or script withdraw-0, same auth pattern as `TransferAct`.
- Reshaping their value into any number of PLB outputs, each at the same payment_credential (PLB) and the same stake_credential pinned by the first PLB input.
- Free choice of datum and `reference_script` on each new PLB output. Unfracking is by design a layout-restructuring action.
- Moving ada freely in/out of PLB UTxOs. Ada is not programmable; the validator's scope is programmable tokens.

**Forbids:**
- Any mint or burn (`tx.mint` must be zero). Unfracking is strictly value-preserving; mint/burn is `issuance_mint`'s job.
- Sending programmable tokens to a non-PLB address — falls out for free from the per-policy equality at PLB + ledger conservation + `mint == 0`.
- Mixing different stake credentials across PLB inputs (or routing a PLB output to a different stake_cred).


lib/types.ak
```diff
@@ -26,6 +26,20 @@ pub type ProgrammableLogicGlobalRedeemer {
...
+  /// one policy stops collateral-damaging unrelated policies that
+  /// happened to share the same UTxO.
+  ///
+  /// No payload: the validator infers everything it needs from the
+  /// transaction shape. Caller proves authorisation by signing for
+  /// (or invoking the script of) every PLB input being unfracked;
+  /// the validator enforces value conservation per policy at the PLB
+  /// boundary and refuses to route tokens to any stake credential the
+  /// caller did not authorise.
+  UnfrackingAct
```
KtorZ, line 50: Haven't looked in depth in the question here but; it might be worth considering delegating the "unfracking" logic into its own script (invoked e.g. via a withdraw-0 script); to avoid increasing the size of the transfer script for something that may only be used rarely. 

This way; the "transfer" script can simply delegate to another script and remain small overall by allowing the unfracking logic to be provided "on demand"

nemo83: At the moment we have:
1. Prog Logic Base  (spend, delegates to Prog Logic Global)
2. Prog Logic Global (Transfer, Third Party Transfer, Unfrack)

The suggestion here is to delegate from PLG to Unfrack Script, which would mean 2 delegations: PLB -> PLG -> Unfrack

It could have made sense remove one indirection and allow PLB -> Unfrack (by setting the script hash another param) but would require cehcking every single time teh redeemer and introducing extra costs.

Doing the double redirection , would keep the PLB code costs at its minimum, no datum nor redeemer parsing required.

nemo83:
Switching back to draft as I'm back to the drawing board for some optimizations.

Kammerlo (approved):
LGTM, but I would still rely on an additional opinion on this one.

fabianbormann (approved):

fabianbormann:
Looks good to me!

PR #80, opened 2026-06-09
fix(issuance_mint): Finding 04 — precise per-redeemer scope delegation

### LINKED ISSUE #56: Escape of Minted Tokens during Third-Party Action
# Finding 04: Escape of Minted Tokens during Third-Party Action

## Summary

`issuance_mint` delegates mint-output custody checks whenever the transaction spends **any** UTxO from `programmable_logic_base`.

That assumption is valid for mint-aware transfer flows, but it is **not** valid for `ThirdPartyAct`. The third-party path only enforces balance preservation for the seized policy referenced by its registry node.

As a result, a transaction can:

1. spend programmable token `A` under `ThirdPartyAct`
2. mint registered token `B` in the same transaction
3. route newly minted `B` to an arbitrary non-PLB address

Both validators accept the transaction.

## Root cause

In the `RefInput` branch, `issuance_mint` decides whether to skip its own output-custody enforcement using this condition:

- **does the transaction spend any input whose payment credential is `programmable_logic_base`?**

If yes, it assumes PLGlobal will validate the rest:

```aiken
or {
  list.any(
    self.inputs,
    fn(input) {
    },
  ),
  validate_mint_outputs(...)?,
}
```

While PLGlobal's `transfer_act` path does enforce custody for all programmable tokens in the transaction, PLGlobal's `third_party_act` path does not.

PLGlobal's `validate_3rd_party` function extracts a **single** seized policy from the registry node and only adjusts accounting for that policy:

```aiken
let
  policy_id,
  third_party_logic,
<- with_key_and_3rd_party_logic(registry_node)

let minted_or_burned = self.mint |> value.tokens(policy_id)
```

This `minted_or_burned` does not include any other programmable tokens minted in the transaction, so PLGlobal does not enforce that they remain in programmable custody.

So when token `B` is minted while `ThirdPartyAct` is operating on token `A`, no core validator enforces that minted `B` remains at programmable custody.

`check_seized_tokens` does not close this gap. The `split_at` equality checks (`input_tokens_after == output_tokens_after`) only compare the seized input against its directly paired output — token `B` is not present in any input (it is freshly minted) so it never appears in `input_tokens_after`. The final output sweep filters by `output.address.payment_credential == prog_logic_cred`, which drops the non-PLB output carrying token `B`. Even if that filter were absent, `value.tokens(policy_id)` scopes to token `A`'s policy and returns empty for any token `B` output. Token `B` contributes nothing to any accumulator in the function.

## Exploit scenario

Assume:

- token `A` is a registered programmable token
- token `B` is another registered programmable token
- the attacker or operator has valid mint authority for token `B`
- the transaction includes a valid `ThirdPartyAct` for token `A`

The attacker builds a transaction with:

- an input spending token `A` from `programmable_logic_base`
- `ThirdPartyAct` withdrawals and redeemer for token `A`
- a mint of token `B` via `issuance_mint` using `RefInput`
- an output sending minted token `B` to a normal non-PLB address

Validation outcome:

- `issuance_mint(B)` sees a PLB spend and delegates instead of checking `B` custody
- `validate_3rd_party(A)` only checks token `A`
- token `B` escapes the programmable shell

## Impact

This breaks the core invariant that registered programmable tokens minted through `issuance_mint` must remain under programmable custody.

A minter with valid mint authority for token `B` can use an unrelated third-party/admin action on token `A` as a false delegation trigger to mint `B` directly to arbitrary addresses.

For regulated assets, this is a programmable-custody escape and compliance bypass.

**Likelihood rationale (Low):** exploitation requires a registered protocol participant — a known, on-chain-identifiable entity — who simultaneously holds mint authority for token `B` and can trigger a `ThirdPartyAct` (via their own token or in collusion with a compliance operator). This is an insider threat scenario: the attacker cannot be anonymous, the exploit is detectable on-chain, and the motivation (bypassing your own token's compliance rules at mint time) is narrow.

## Proof of exploit

Added a cross-validator exploit test:

- `validators/programmable_logic_global.test.ak`

Test name:

- `third_party_act_on_token_a_plus_issuance_mint_of_token_b_escape_succeeds`

The test:

- spends token `A` from PLB
- invokes `ThirdPartyAct` for token `A`
- mints token `B`
- routes minted token `B` to a non-PLB wallet output
- calls both `issuance_mint.mint(...)` and `programmable_logic_global.withdraw(...)`

The exploit shape is accepted by both validator calls.

## Recommended remediation

**Option 1**

Modify `issuance_mint`'s delegation condition to check the PLGlobal redeemer:

- Only allow `issuance_mint` to delegate its output-custody checks when the PLGlobal's redeemer is `TransferAct`.
- Otherwise, if the redeemer is `ThirdPartyAct`, keep `issuance_mint`'s output-custody checks active.

**Option 2**

Modify `ThirdPartyAct` to enforce programmable-custody/accounting for every programmable policy minted in the transaction, not just the seized policy referenced by its registry node:

- Extend `ThirdPartyAct` to enforce programmable-custody/accounting for every programmable policy minted in the transaction.

## Short conclusion

This is a real cross-validator composition flaw:

- `issuance_mint` trusts PLGlobal too broadly
- `ThirdPartyAct` does not uphold that trust for other minted policies

The result is that a mint of token `B` can escape programmable custody when combined with a third-party action on token `A`.

nemo83:
Fixes #56 

## Summary

Closes audit Finding 04 (`audit/04-third-party-act-can-falsely-satisfy-issuance-delegation.md`) by tightening `issuance_mint`'s delegation decision to match PLGlobal's actual per-redeemer scope.

## The change

Issuance mint delegates prog token balance invariance checks to PLG if and only if the Transfer or Third Party Transfer list the prog token registry token in their redeemer. Checks minting balance otherwise.


fabianbormann (approved):

fabianbormann:
Looks good to me. Sorry for the delay in reviewing

PR #81, opened 2026-06-11
feat: added in place update for registry node fields

### LINKED ISSUE #75: Registry Design Limitation
# Registry Design Limitation

## Description

The current registry design does not allow for updating registry nodes or de-registering them. This means that once a registry node is added to the list, it cannot be removed or updated.

The current registry node structure is as follows:

```rs
pub type RegistryNode {
  /// The key (currency symbol) of the programmable token policy
  key: ByteArray,
  /// The next key in lexicographical order (for linked list traversal)
  next: ByteArray,
  /// The transfer logic script credential that validates token transfers
  transfer_logic_script: Credential,
  /// The third party transfer logic script credential that handles admin actions (seizure, etc.)
  third_party_transfer_logic_script: Credential,
  /// Optional global state NFT currency symbol for this policy
  global_state_cs: ByteArray,
}
```

The `transfer_logic_script` and `third_party_transfer_logic_script` fields are crucial for the functionality of the registry nodes, as they define the logic for validating token transfers and handling admin actions. If there are bugs or vulnerabilities in the initial implementation of these scripts, there is currently no way to update them, which could lead to potential security issues.

## Current Limitations

1. No way to fix issues: If there are bugs or vulnerabilities in the initial implementation of the transfer logic script or the third-party transfer logic script, there is currently no way to update them, which could lead to potential security issues.
2. No way to add features: If the issuer wants to add new features to the transfer logic or the third-party transfer logic, there is currently no way to do that, which could limit the functionality of the registry nodes.
3. No way to de-register tokens and move out of the CIP-113: If a stable-coin issuer decides that they want to move out of this cip-113 compliant registry, there is currently no way to de-register their token from the registry.

In the current design, if a token needs to be updated, there is no direct way to do that, instead, the issuer would have to create a new token, register it, and then migrate all the users to the new token, which could be a complex and error-prone process. Additionally, the old token would still be in the registry, and its minting logic would still be active, which could lead to confusion and potential security issues.

## Recommendations

The main question is who should have the authority to perform registry-related actions, such as registering a programmable token, updating the token's registry node, and de-registering the token. One potential solution is to use the `issuanceLogicScript` as the main component in authorizing registry-related actions.

We recommend defining a `RegistryLifecycleAuthority` witness, which proves that the token-specific issuance authority has authorized a registry lifecycle action (register, update, de-register).

The `RegistryLifecycleAuthority` can be defined as follows:

For a programmable token policy `P`, `RegistryLifecycleAuthority` means:

1. P is the official issuanceMintingPolicy instance, derived from the provided issuanceLogicScript.
2. The corresponding issuanceLogicScript is executed through withdraw-zero.
3. tx.mint has no entries under policy P.

Note: the 3rd condition (`tx.mint` has no entries under policy P) is used to separate the authority of the `RegistryLifecycleAuthority` from the authority of issuing new tokens, as the `RegistryLifecycleAuthority` should only have the authority to perform registry-related actions, and not the authority to mint new tokens. Similarly, issuing new tokens should not grant the authority to perform registry-related actions.

The `RegistryLifecycleAuthority` can be implemented as follows (conceptually):

> **Implementation note:** If `issuanceLogicScript` is reused as the registry lifecycle authority, the registry must not merely check that some withdrawal script was executed. It must prove that the executed `issuanceLogicScript` is the same credential that parameterizes the official `issuanceMintingPolicy` instance for policy `P`. The exact parameter bytes used to reconstruct the issuance policy hash are serialization-sensitive and should be confirmed with golden tests against the compiled/applied issuance policy.

```rs
fn credential_to_bytes(cred: Credential) -> ByteArray {
  builtin.serialise_data(as_data(cred))
}

pub fn has_registry_lifecycle_authority(
  token_policy: ByteArray,
  issuance_logic_cred: Credential,
  all_mint_entries: Value,
  withdrawals: Pairs<Credential, Lovelace>,
  prefix: ByteArray,
  postfix: ByteArray,
) -> Bool {
  let param_bytes = credential_to_bytes(issuance_logic_cred)
  let computed_policy = apply_hashed_parameter(prefix, postfix, param_bytes)

  and {
    // Check if the issuanceLogicScript is executed via withdraw-zero
    list.any(
      withdrawals,
      fn(wdrl) -> Bool {
        let Pair(cred, _amount) = wdrl
        cred == issuance_logic_cred
      },
    ),
    // Check if tx.mint has no entries under policy P
    dict.is_empty(assets.tokens(all_mint_entries, token_policy)),
    // Check if the issuanceMintingPolicy instance is derived from the provided issuanceLogicScript
    computed_policy == token_policy,
  }
}
```

The `RegistryLifecycleAuthority` can then be used to authorize the following registry-related actions:

1. Register a programmable token: The `RegistryLifecycleAuthority` can be used to authorize the registration of a new programmable token in the registry.
2. Update the token's registry node: The `RegistryLifecycleAuthority` can be used to authorize updates to the transfer logic script, third-party transfer logic script, or global state of an existing registry node.
3. De-register the token: The `RegistryLifecycleAuthority` can be used to authorize the de-registration of a token from the registry.

- Note that the de-registration path should not allow for the complete deletion of the registry node, as this would lead to the following potential security issue:
  - If a registry node is deleted (completely removed from the registry), it could potentially be re-registered, which would create an escape window for the tokens to be moved outside the PLB.
  - Attack scenario: The token issuer de-registers their token from the registry, then tokens are transferred or minted outside the PLB, and then the issuer re-registers the same token back to the registry. This would allow the tokens to effectively escape the PLB, which would create a major security issue.

- Instead, the de-registration path should clearly mark the registry node as de-registered, which would prevent re-registration and ensure that the token is effectively removed from the registry while still maintaining a record of its existence. This could be done by adding a "de-registration" flag in the registry node, which would indicate that the node is no longer active and should not be used for any registry-related actions. Note also that the de-registration path should be irreversible, meaning that once a registry node is marked as de-registered, it cannot be re-activated or re-registered.

## Summary

The current registry design does not allow for updating registry nodes or de-registering them, which could lead to potential security issues and limit the functionality of the registry. We recommend defining a `RegistryLifecycleAuthority` that would have the authority to perform registry-related actions, such as registering a programmable token, updating the token's registry node, and de-registering the token. This would allow for a more flexible and secure registry design, as it would allow for updating registry nodes and de-registering them when necessary.

Kammerlo:
Implementing the path to update a registry node. By spending an utxo and creating a new one.
With this change it is possible to update:
- transfer_logic_script
- third_party_logic_script
- global_state_cs
And enforcing fixed: key, next and minting_logic_script 

Implementation of: https://github.com/cardano-foundation/cip113-programmable-tokens/issues/75

### DIFF
--- a/lib/linked_list.ak
+++ b/lib/linked_list.ak
@@ -169,3 +169,25 @@ pub fn validate_mint(mint: Value, policy_id: PolicyId) -> AssetName {
     |> dict.to_pairs()
   asset_name
 }
+
+// True iff `new` equals `old` except possibly the mutable fields: 
+// transfer_logic_script, third_party_transfer_logic_script, and global_state_cs.
+// All other fields must still be well-formed. The record freezes key, next, and minting_logic_script
+pub fn is_field_updated_registry_node(
+  new: RegistryNode,
+  old: RegistryNode,
+) -> Bool {
+  and {
+    new == RegistryNode {
+      ..old,
+      transfer_logic_script: new.transfer_logic_script,
+      third_party_transfer_logic_script: new.third_party_transfer_logic_script,
+      global_state_cs: new.global_state_cs,
+    },
+    is_28_byte_credential(new.transfer_logic_script)?,
+    is_28_byte_credential(new.third_party_transfer_logic_script)?,
+    (bytearray.length(new.global_state_cs) == 0 || is_28_bytes(
+      new.global_state_cs,
+    ))?,
+  }
+}
--- a/lib/linked_list.ak
+++ b/lib/linked_list.ak
@@ -179,7 +179,9 @@ pub fn is_field_updated_registry_node(
 ) -> Bool {
   and {
     new == RegistryNode {
-      ..old,
+      key: old.key,
+      next: old.next,
+      minting_logic_script: old.minting_logic_script,
       transfer_logic_script: new.transfer_logic_script,
       third_party_transfer_logic_script: new.third_party_transfer_logic_script,
       global_state_cs: new.global_state_cs,
--- a/validators/registry_spend.ak
+++ b/validators/registry_spend.ak
@@ -68,7 +68,7 @@ validator registry_spend(protocol_params_cs: PolicyId) {
         when old_node.minting_logic_script is {
           Script(_) ->
             pairs.has_key(self.withdrawals, old_node.minting_logic_script)
-          VerificationKey(pkh) -> list.has(self.extra_signatories, pkh)
+          VerificationKey(_) -> False
         }
       }
     }
--- a/validators/registry_spend.ak
+++ b/validators/registry_spend.ak
@@ -10,7 +10,7 @@ use cardano/address.{Script, VerificationKey}
 // the registry structure is being properly maintained.
 
 use cardano/assets.{PolicyId}
-use cardano/transaction.{Transaction}
+use cardano/transaction.{OutputReference, Transaction, find_input}
 use linked_list.{is_field_updated_registry_node, validate_directory_node_output}
 use programmable_logic/params.{ProgrammableLogicGlobalParams}
 use registry_node.{RegistryNode}
@@ -18,9 +18,9 @@ use utils.{expect_inline_datum, has_currency_symbol}
 
 validator registry_spend(protocol_params_cs: PolicyId) {
   spend(
-    _datum: Option<Data>,
+    datum: Option<RegistryNode>,
     _redeemer: Data,
-    _own_ref: Data,
+    own_ref: OutputReference,
     self: Transaction,
   ) {
     trace @"Starting registry_spend validation"
@@ -50,11 +50,8 @@ validator registry_spend(protocol_params_cs: PolicyId) {
       }
       // No registry node NFTs being minted or burned means we are updating a node
       None -> {
-        let own_input = list.expect_find(
-          self.inputs,
-          fn(i) { has_currency_symbol(i.output.value, registry_node_cs) },
-        )
-        expect old_node: RegistryNode = expect_inline_datum(own_input.output)
+        expect Some(own_input) = find_input(self.inputs, own_ref)
+        expect Some(old_node) = datum
         expect [cont] = list.filter(
           self.outputs,
           fn(o) { has_currency_symbol(o.value, registry_node_cs) },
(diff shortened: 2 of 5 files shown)


lib/linked_list.ak
```diff
@@ -169,3 +169,25 @@ pub fn validate_mint(mint: Value, policy_id: PolicyId) -> AssetName {
...
+// True iff `new` equals `old` except possibly the mutable fields: 
+// transfer_logic_script, third_party_transfer_logic_script, and global_state_cs.
+// All other fields must still be well-formed. The record freezes key, next, and minting_logic_script
+pub fn is_field_updated_registry_node(
+  new: RegistryNode,
+  old: RegistryNode,
+) -> Bool {
+  and {
+    new == RegistryNode {
+      ..old,
```
nemo83: nitpick, does spread operator cost more than just slap that 2 or 3 missing props?

validators/registry_spend.ak
```diff
@@ -34,12 +38,40 @@ validator registry_spend(protocol_params_cs: PolicyId) {
...
+      Some(minted_assets) -> {
+        expect [Pair(_, amount)] = dict.to_pairs(minted_assets)
+        amount > 0
+      }
+      // No registry node NFTs being minted or burned means we are updating a node
+      None -> {
+        let own_input = list.expect_find(
+          self.inputs,
+          fn(i) { has_currency_symbol(i.output.value, registry_node_cs) },
+        )
```
nemo83: In general

1. replace w/ https://aiken-lang.github.io/stdlib/v3.1.0/cardano/transaction.html#find_input 
2. then expect assets.has_nft_strict(input.output.value, registry_node_cs)

Becasue in your find thats' any input, not guarateed to be your own.

BUT

if you trust the previous steps (ie. init + insert), than you can't have an input w/o the nft. So you could skip everything and just extract the datum from the spend signature

expect Some(old_node_datum) = _datum
expect old_node: RegistryNode =  old_node_datum

and skip all the discovery....

nemo83 (approved):
couple of minor notes

nemo83 (changes requested):
I think you forgot to ensure no programmable tokens are minted as part of this TX, which I don't think it's a vulnerablity per se, but better to not allow things that are not supposed to happen anyway

nemo83 (approved):
LGTM

Issue #73, opened 2026-05-26
3rd Party Act Path Scope is Not Clearly Defined


nemo83:
# Finding 18: 3rd Party Act Path Scope is Not Clearly Defined

## Description

In the current implementation of the `ThirdPartyAct` path in the PLGlobal, there is a lack of clarity regarding the scope of actions that an admin can perform.

The current implementation of the `ThirdPartyAct` path allows for a wide range of actions that an admin can perform, without clear guidelines or restrictions. This can lead to potential misuse of the admin privileges, or unintended consequences for users interacting with the protocol.

Currently, the `ThirdPartyAct` gives the admin control over all UTxOs that holds the administered token, including the UTxOs that belong to users, DeFi protocols, DEXs, and other applications. This means that the admin can perform actions on these UTxOs, such as consuming them, which can have significant implications for users and the overall ecosystem.

## Impact

One of the main impacts of this issue is related to the usability and security of DeFi protocols. While the `ThirdPartyAct` path is intended for administrative purposes, it can be used to perform actions that may be considered "dangerous" or beyond the scope of typical administrative functions. For example, in DeFi applications, the `ThirdPartyAct` path can be used to consume collateral tokens, or may be able to force liquidation of positions, which can have significant financial implications for users and may affect the usability and security of DeFi protocols.

Users that want to interact with DeFi applications may be hesitant to do so if they know that an admin has the power to perform actions that could negatively impact their assets or positions. This can lead to a lack of trust in the protocol and may discourage users from participating in the DeFi ecosystem. For example, if an admin can seize collateral tokens, a lender may be hesitant to lend assets on the platform, as they may fear that the collateral that covers their loan could be seized by the admin. This can lead to a decrease in liquidity and overall participation in the DeFi protocol, which can ultimately harm the growth and success of both the DeFi protocol and the CIP-0113 standard.

## Recommendation

1. Clearly define the scope of actions that the admin can perform through the `ThirdPartyAct` path, and ensure that these actions are limited to necessary administrative functions. This will help to prevent misuse of admin privileges and reduce the risk of unintended consequences for protocols implementing the CIP-0113 standard.
2. After clearly defining the scope of admin actions, document these actions in the protocol's documentation to provide transparency to users and build trust in the protocol. This will help users understand the limitations of admin actions and make informed decisions when interacting with the protocol.
[...]

Issue #71, opened 2026-05-26
Token UTxO Fragmentation Causes Non-Atomic Seizure


nemo83:
# Finding 16: Token UTxO Fragmentation Causes Non-Atomic Seizure

## Description

A token holder can spread their balance across many UTxOs. Because `check_seized_tokens` pairs PLB inputs with outputs positionally and Cardano transactions are bounded by size and execution-unit limits, only a limited number of UTxOs can be processed per transaction. 
When a compliance operator needs to seize the holder's entire balance, too many UTxOs may exceed what fits in a single transaction, requiring multiple transactions to complete the full seizure.

## Scenario

**Complete seizure of a highly fragmented holder**

- A holder with more UTxOs than fit in a single transaction requires multiple seizure transactions to seize the full balance, these transactions may be included in the same block or across separate blocks.
- If the transactions span multiple blocks, unprocessed UTxOs remain live and spendable between blocks.
- Each transaction may succeed or fail independently, making partial completion a realistic outcome.

## Impact

When seizing all tokens from a highly fragmented holder, the operation may be only partially successful if not all transactions land in the same block. Some UTxOs may be seized while others remain accessible to the holder during the multi-transaction window. This weakens compliance guarantees compared to an atomic, all-or-nothing operation.

If the token requires atomic full seizure, a freeze mechanism must be in place as a prerequisite.

## Recommendation

Document this limitation explicitly in the CIP-113 specification so that compliance operators are aware of the multi-transaction nature of seizure for fragmented holders.

Third-party substandards that require strong seizure guarantees should include a freeze mechanism as a prerequisite step. Freezing must block all spends from the target's credentials before any seizure transaction is submitted, eliminating the window in which remaining UTxOs can be moved. The substandard specification should state this as a required guarantee rather than leaving it as an implementation detail.

nemo83:
Documented (no code change). Seizure is per-UTxO and fragmentation is inherent to the eUTxO model — not a framework defect; full seizure of a fragmented balance may need multiple transactions. Mitigated (not eliminated) by the holder-driven Unfracking action. See `documentation/03-CONTROL-SCOPE-AND-ADMIN-AUTHORITY.md` §3.4.

Issue #67, opened 2026-05-20
`ThirdPartyAct` Cannot Atomically Operate Across Multiple Token Policies


nemo83:
# Finding 15: `ThirdPartyAct` Cannot Atomically Operate Across Multiple Token Policies

## Description

`validate_3rd_party` resolves exactly one `registry_node_idx` to one `policy_id` per invocation. Because a withdrawal script can only be invoked once per transaction, only one token policy can be processed in a single `ThirdPartyAct` transaction. This prevents any compliance operation that must atomically involve tokens from two or more distinct policies.

## Scenario

**Token conversion / replacement**

A compliance operator needs to atomically retire two legacy token policies (A and B) and issue a replacement token (C), such as during a token migration, a corporate restructuring, or a regulatory-mandated instrument replacement.

- Burning policy A tokens and policy B tokens must occur in the same transaction as minting policy C tokens to preserve atomicity and auditability.
- Under the current design, only one of A or B can be processed per `ThirdPartyAct` invocation, making a true atomic swap impossible.
- The operator must accept a multi-step process with an exposure window between steps, or rely on off-chain coordination to approximate atomicity.

## Impact

Operations involving more than one token policy require multiple sequential transactions rather than a single atomic one, adding operational steps and leaving intermediate states briefly visible on-chain.

## Recommendation

Extend the redeemer to accept a list of policy references:

```aiken
ThirdPartyAct { actions: List<(registry_node_idx: Int, outputs_start_idx: Int)> }
```

Each entry targets one policy; all are validated in a single transaction, bounded only by execution-unit and size limits. This enables atomic multi-policy compliance operations without a protocol redesign.

nemo83:
Won't-fix. Atomic multi-policy ThirdPartyAct was prototyped but deliberately not adopted: making the path multi-policy imposed a significant execution-cost and script-size tax on the common single-policy case. Documented as a permanent limitation in `documentation/03-CONTROL-SCOPE-AND-ADMIN-AUTHORITY.md` §3.1; compliance operations spanning multiple policies use sequential transactions.

Issue #64, opened 2026-05-20
ThirdPartyAct UTxO Contamination


nemo83:
# Finding 12: ThirdPartyAct UTxO Contamination

## Description

In the `ThirdPartyAct` path of the `programmable_logic_global` validator, the `check_seized_tokens` function enforces that the tokens affected by the 3rd-party action (referred to as "seized tokens") must change, but only those tokens. This means that all non-target-policy assets must remain unchanged, and only the target policy tokens can be modified.

However, the wording "Seized tokens" is misleading, because technically, the logic allows changing these tokens' amounts in various ways, other than just seizing them (e.g. adding more tokens to existing tokens, or adding new tokens that were not present before).

For example, the 3rd-party action can be used to add new tokens (i.e. inject new tokens into the `value` of the output UTxOs). As long as the total amount of the target policy tokens is preserved across inputs (including minting or burning effects) and outputs, the transaction can still be valid. So, the 3rd-party action can be used to manipulate tokens in various ways, not just seizing them, which can lead to a wide range of potential attack vectors if not properly understood and mitigated.

Moreover, the `ThirdPartyAct` path can include any UTxO from the CIP-113's PLB address as input in the transaction. There is no restriction on which UTxOs can be included in the transaction.

Thus, a malicious programmable token’s `ThirdPartyAct` can contaminate arbitrary PLB UTxOs that did not previously contain that token, because the paired input is not required to already hold the acted-on policy.

## Relevant Code Snippet

```rs
/// Analyze inputs and outputs affected by the 3rd-party action (grossly named *seized_tokens* here).
fn check_seized_tokens(..) -> Bool {
    ...
        // Seized tokens must change, but only them.
        expect and {
            (input_tokens_before == output_tokens_before)?,
            (input_tokens_after == output_tokens_after)?,
            (input_tokens_at != output_tokens_at)?,
          }
    ...
}
```

Here `input_tokens_at` and `output_tokens_at` refer to the amounts of the target policy tokens in the inputs and outputs affected by the 3rd-party action, while `input_tokens_before`, `output_tokens_before`, `input_tokens_after`, and `output_tokens_after` refer to the amounts of non-target-policy assets in the inputs and outputs before and after the 3rd-party action.

[...]

PR #82, opened 2026-06-22
feat: control & admin scope — protected prefixes + scope documentation (Findings 01, 15, 16, 18)

nemo83:
## Summary

Resolves the **scope-of-programmable-control** and **admin / `ThirdPartyAct` scope** audit findings: adds the protected-prefix mechanism to the registry and the third-party path, and the documentation that defines exactly what the framework and an administrator can and cannot do to a holder's tokens.

Builds on #81 (registry-node in-place update), which is merged into this branch; the append-only enforcement for `protected_prefixes` lives in `is_field_updated_registry_node`.

## Code — Finding 18 (#73), with Finding 12 (#64) hardening

- **`RegistryNode.protected_prefixes`** — an issuer-declared, append-only, strictly-ascending list of 4-byte CIP-67 asset-name label prefixes. `ThirdPartyAct` may **neither extract nor burn** tokens whose label is protected ("preserve, not fail"): only the non-protected remainder of the same UTxO is seizable.
- Validation is a **single pass** (well-formedness + append-only) shared by the insert and update paths; an **empty-list fast path** keeps the common (no-protected-tokens) case close to baseline cost.
- **`ThirdPartyAct` must change the subject balance** — re-introduces the `input_tokens_at != output_tokens_at` check that Finding 12 recommended removing as redundant; kept deliberately as an **anti-DDoS** guard so an admin cannot force-respend a holder's UTxO with no economic change.

## Docs — Findings 01 (#53), 15 (#67), 16 (#71), 18 (#73)

New `documentation/03-CONTROL-SCOPE-AND-ADMIN-AUTHORITY.md`:

- **F01** — scope of programmable control: metadata/royalty (CIP-68/102) are **out of the control surface**, handled by CIP-68/102-aware substandards. Companion assets stay in the PLB (no escape); transferring them is allowed via the substandard's transfer logic, while `ThirdPartyAct` is forbidden from interfering (protected prefixes).
- **F18** — admin scope: structural guarantees, protected prefixes, the **freeze-vs-extract** asymmetry, **holder scope** (vkey vs script) as a substandard decision, and the DeFi-aware **allowlist / consent** reference patterns.
- **F15** — exactly **one policy per `ThirdPartyAct`** (deliberate, permanent limitation — won't-fix).
- **F16** — seizure is **per-UTxO**; fragmentation is not prevented (inherent to eUTxO, mitigated by Unfracking).
- Registry-node update-authority transparency (retroactive effect, credential-type flips; the registry node is the live source of truth).

Synced `02-ARCHITECTURE.md` and `09-DEVELOPING-SUBSTANDARDS.md`: the `RegistryNode` shape, `ThirdPartyAct` described as a general administrative/compliance action (not only seizure), and the companion-asset + holder-scope guidance.


fabianbormann (approved):

fabianbormann:
Looks good to me ! 🔥

Kammerlo (approved):

Issue #96, opened 2026-07-27
[AL] ThirdPartyAct pair check forces exact lovelace equality; a min-ADA parameter increase makes seizure unsatisfiable


nemo83:
## Summary

Reported by Phil di Sarro (Anastasia Labs), 2026-07-27. The ThirdPartyAct per-pair value check compares full `Value`s, so the paired continuing output must carry exactly the input's lovelace. A protocol parameter update that raises the min-ADA result for an existing PLB UTxO's shape — even by 1 lovelace — makes every third-party action over that UTxO fail phase-1 validation, permanently, with no admin-side remediation.

## Mechanism

- `validate_io_constraints_and_balance` (`validators/programmable_logic/third_party.ak`) compares each paired PLB input/output via `assets.split_at(value, policy_id)` applied to the **full** `Value`. The ada entry (empty policy id) sorts first and always lands in the `before` half, so `input_tokens_before == output_tokens_before` enforces exact lovelace equality in addition to the intended byte-identity of non-acted-on policies.
- The transfer and unfracking paths strip the ada entry (`from_value |> list.tail` in `lib/assets.ak` / `validators/programmable_logic/unfracking.ak`) before comparing; the ThirdPartyAct pairing is the only path that includes it.
- A UTxO's lovelace is fixed at creation; the ledger's min-ADA check (`coinsPerUTxOByte`) is evaluated at spend time. After a parameter increase, the forced paired output (same address, datum, reference script; acted-on tokens removed) can fall below the current min-ADA threshold. The equality forbids topping it up, so no satisfying transaction exists.
- The paired output only ever *loses* value relative to the input, so the failure is triggered exclusively by the parameter change — UTxOs funded at exact min-ADA under the old parameters are the affected population, which at the time of a change is potentially every existing PLB UTxO.

## Impact

Loss of the compliance guarantee ThirdPartyAct exists to provide. Holders are unaffected (TransferAct permits free restructuring and does not constrain lovelace), but ThirdPartyAct is precisely the path that must succeed **without** holder cooperation: after a min-ADA-relevant parameter change, an uncooperative (e.g. sanctioned) holder's pre-existing UTxOs become unseizable by simply never being spent. It additionally creates an incentive to hold programmable tokens in exact-min-ADA UTxOs in anticipation of such a change.

## Proposed remediation

[...]

Issue #94, opened 2026-07-23
[AL] protected_prefixes cost amplification enables registry-wide insert bricking


nemo83:
## Summary

Reported by Phil di Sarro (Anastasia Labs), 2026-07-20,. The unbounded `RegistryNode.protected_prefixes` field lets an attacker inflate the cost of future registry insertions past execution limits, bricking registration for the whole list.

## Mechanism

- `RegistryNode.protected_prefixes` (`lib/registry_node.ak`) is a `List<ByteArray>` of 4-byte CIP-67 prefixes, strictly ascending, append-only. There is **no length cap** — bounded only by tx/datum limits (~16 KB).
- Every insert must CBOR-decode the covering node, re-validate both outputs, and run a full record equality over the entire node (`lib/linked_list.ak`) — all O(|protected_prefixes|). The inserter also reconstructs and pays min-ADA for the covering node's bloated datum.
- The attack **never touches an existing node**. A registrant's `key` is the policy id derived from its `minting_logic_script`, so list position is grindable: mine policy ids adjacent to every existing node and insert attacker-owned, maximally-bloated nodes there. Every future insertion then lands on a bloated covering node and cannot fit within execution limits.
- The append-only + issuer-only constraints on the field do **not** mitigate this, since the attacker only ever bloats their own nodes.

## Impact

Permanent denial of registration (registry bricking) at attacker-chosen positions, up to the entire list. AL has a reproducible end-to-end PoC.

## Reporter's recommendation

Remove `protected_prefixes` from core (the field, its accessors, insert/update validation, and the ThirdPartyAct consumption). The registered `third_party_transfer_logic_script` is already invoked on every third-party action and can express royalty/metadata prefix protection itself, so the feature is preserved at the substandard layer while the core stays unopinionated. Phil also noted this aligns with his original issue-#53 advice to keep metadata/royalty control out of the standard's control surface.

_Remediation is under internal discussion; this issue records the report as received._

PR #100, opened 2026-08-06
docs: sync documentation with #97 and #98

jorbuedo:
Neither #97 (remove `protected_prefixes` + lovelace ratchet) nor #98 (unfracking v2) touched any documentation, and the docs have not been updated since #92. The result is that **seven files describe a `RegistryNode` field that no longer exists, and none describe the field that replaced it.**

Docs only — no validator changes. `aiken fmt --check` and `aiken check` pass unchanged.

### The part that actively misleads integrators

`RegistryNode` is still 7 fields, but the layout moved:

```
before #97/#98                         current main
5  global_state_cs                     5  unfracking_logic_script
6  protected_prefixes                  6  global_state_cs
```

The datum blocks in `02-ARCHITECTURE.md`, `09-DEVELOPING-SUBSTANDARDS.md`, `cip113-api-changes-post-audit.md` and `cip113-substandard-upgrade-guide.md` all still show the pre-#97 order. Anyone following them writes a `Credential` where the validator reads a `ByteArray`.

All four now match `lib/registry_node.ak`. Because *both* revisions have 7 fields, I also added an explicit warning to §2 of the API-changes doc that field **count** does not identify the revision — parsers have to match on positions.

### `protected_prefixes`

Rewritten as a substandard responsibility, following the rationale in #97: the framework guarantees the subject policy's `third_party_transfer_logic_script` runs on every `ThirdPartyAct`, so a substandard can protect its own CIP-67 companion labels.

Where the guidance is *reversed* rather than merely changed, I kept a short "changed in #97" note pointing at #94 — a substandard author who already declared prefixes is silently unprotected now and needs to reimplement it in their third-party validator. Finding 18 is marked **reverted** in the findings table rather than deleted, so the audit trail stays readable.

### Also swept up

- **Lovelace ratchet** (#96) — §10 still described per-pair lovelace as byte-equal; it is now `output >= input`. Documented with the reason for the inequality (a min-ADA rise must be absorbable, and `>=` only so an admin can't drain ada).
- **`UnfrackingRedeemer` + default-deny hook** — §9 gains the redeemer shape, the four-withdrawal composition, and the fact that an unset `unfracking_logic_script` forbids unfracking outright. Worth surfacing to builders before submission rather than at validation time.
- **`unfracking_logic_script` added to every mutable-fields list** (§7, the upgrade guide's mutability table, `03-CONTROL-SCOPE`, and the integration guide's caching warning), including that it can be unset back to `empty_vkey`.
- **`RegistryInsert` `mode` tag** — the upgrade guide's verification checklist still asked integrators to "set `mode` correctly" after #92 removed it.
- Baseline commit in the API-changes doc updated `3a6d8d6` → `f16c857`, with the intervening PRs listed.

I verified each datum block field-by-field against `lib/registry_node.ak` rather than transcribing between docs. Happy to split this per-file or drop any section if you'd rather word the substandard-migration guidance yourselves.


nemo83:
Hi, sorry for the delay and the PR, as the code is currently changing, we will process this PR as soon as the code stabilises.

Thanks for your contribution and patience

nemo83:
this is now fixed by #139 thanks for reporting the issue.

PR #104, opened 2026-08-18
Harden in-place upgrade path: validate credential well-formedness in upgrade datum

nemo83:
## Summary

Hardening for the in-place upgrade path. The coordination-spend upgrade rail
validates the continuing params datum with a type/arity cast only
(`expect new_params: ProgrammableLogicGlobalParams = cont_datum`). A `Credential`
carries no length obligation, so an upgrade could write an empty or wrong-length
credential into a mutable authority field and the cast still succeeded. Only
`prog_logic_cred` and `registry_node_cs` were frozen; the mutable authority
credentials went unchecked.

An unsatisfiable credential can never equal any withdrawal key, so
`has_key_or_fail` can never match — permanently bricking the field it gates:

- `prog_logic_global_cred` — PLB's per-spend gate → **every token frozen**.
- `upgrade_logic_cred` — the authority for future upgrades → **coordination UTxO frozen**.
- `unfracking_cred` — the unfracking gate → unfracking disabled.

No adversary is required: reaching for the `empty_vkey` ("unset") idiom on the
wrong field is enough. These fields freeze at bootstrap, so the mistake is
permanent.

## Fix

Add a 28-byte well-formedness check (`is_28_byte_credential`, already used on the
registry path in `lib/linked_list.ak`) to the three mutable authority
credentials in `coordination_spend`.

**Scope of the check:** this validates the *shape/length* of the credential
only. It cannot verify that a credential resolves to a deployed, executable
script — that is out of reach on-chain — but it rejects the wrong-length
credentials that are unconditionally unsatisfiable, which is the class that
silently bricks the protocol. The frozen anchors are already well-formed by
construction (equal to the mint-validated old datum), so only the mutable fields
are checked.

## Tests

`validators/coordination_spend.test.ak`, new "Credential well-formedness" section:

- `coord_rejects_unsatisfiable_plg_cred` — upgrade sets `prog_logic_global_cred`
  to `empty_vkey`; now rejected (was accepted before this change).
- `coord_rejects_unsatisfiable_upgrade_cred` — same for `upgrade_logic_cred`.

Existing positive controls (`coord_upgrade_succeeds`,
`coord_authority_handover_succeeds`) confirm well-formed 28-byte upgrades are
still accepted. Full suite green (682/682).

## Note

Minor hardening; low urgency (upgrade authority is trusted). Opened against the
in-place upgradability branch and may accumulate a few more small upgrade-path
hardening fixes before it merges.

### DIFF
--- a/validators/coordination_spend.ak
+++ b/validators/coordination_spend.ak
@@ -39,6 +39,7 @@
 use aiken/collection/list
 use cardano/assets
 use cardano/transaction.{InlineDatum, OutputReference, Transaction, find_input}
+use linked_list.{is_28_byte_credential}
 use pairs
 use programmable_logic/params.{ProgrammableLogicGlobalParams}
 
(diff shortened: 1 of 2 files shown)


nemo83:
Superseded by #105, which contains the same `coordination_spend` credential well-formedness fix plus additional work (the registry-node UPDATE-branch test fix and a CI assertions run). Closing to avoid a duplicate change on `coordination_spend.ak`; the fix lives on in #105.

PR #110, opened 2026-08-20
Dissolve PLG: PLB dispatches directly to transfer / third_party / unfracking

nemo83:
## Summary

Dissolves the `programmable_logic_global` (PLG) coordinator. `programmable_logic_base` (PLB) becomes a **dispatcher** that trampolines straight to one of three standalone withdraw-0 validators — **`transfer`** (PLG renamed, transfer invariants only), **`third_party`** (new: seize / clawback / freeze), **`unfracking`** (existing, no longer gated through PLG) — and no delegate is ever reached through another. Ports and extends the design of the wsc-poc `perf/seize-optimization` branch (which split seize out but kept the "global" name and has no unfracking).

Cut from and targeting `feat/upgradability-in-place`. **Breaking (pre-mint) redeemer + validator-set change** — intentional.

## Why

PLG is a reference script paid for (as a size-proportional fee) by **every transfer, forever**. The seize logic is heavy but rare. Hosting it separately keeps those bytes off the transfer hot path, and — because a seize tx no longer loads PLG at all — off the seize path too.

**Measured reference-script footprint per tx** (`aiken build`, this branch vs base) — every transaction kind now loads PLB + exactly one delegate:

| path | before | after | Δ |
|---|---|---|---|
| transfer (PLB + transfer) | 3659 B | 3045 B | **−614 B** |
| seize (PLB + third_party) | 3659 B | 2674 B | **−985 B** |
| unfracking (PLB + unfracking) | 5491 B | 2700 B | **−2791 B** |

The transfer script itself: 3163 → 2177 B (**−31%**); PLB 496 → 868 B (three-arm dispatch). Execution CPU is ~neutral: PLB gains a 2-arm dispatch (~4.2M cpu fixed at withdrawal position 0) and resolves the delegate by a direct `list.expect_at` — O(`wdrl_idx`) cell drops, **not O(1)**, but with no credential comparison on the way, where the old scan compared at every entry. Measured in `wdrl_idx_cost.test.ak` (widths 1–16, positions 0–15): ~1.1M vs ~2.7M cpu per position, break-even at position ~3, ~19M cpu saved at width 16 / position 15.

## What changed (by phase / commit)

- **P0** — `third_party_cred` added to the protocol-params datum (appended as field 5 at this phase; **field 3 in the final layout** after the `5c111f9` reorder below; swappable in place; `coordination_spend` 28-byte-guards it).
- **P1** — new `validators/third_party.ak` (withdraw-0), redeemer `ThirdPartyRedeemer { params_idx, registry_node_idx, outputs_start_idx }`, reusing `validate_3rd_party`.
- **P2** — PLB redeemer → `BaseSpendRedeemer { SpendViaGlobal | SpendViaThirdParty }`, each `{ params_idx, wdrl_idx }`; withdrawal scan → **witnessed index** (direct `list.expect_at`, see above).
- **P3** — heavy `validate_3rd_party` removed from PLG; 26 third-party tests relocated to `third_party.test.ak`; benchmarks re-pointed.
- **P4** — `ThirdPartyAct` constructor removed (`ProgrammableLogicGlobalRedeemer = TransferAct | UnfrackingAct`; `UnfrackingAct` constr index 2→1); **`issuance_mint` mint-custody delegation re-pointed** to recognise coverage from the `third_party` validator's redeemer (Finding 04 precise delegation preserved).
- **P5** — before/after size record in `benchmarks.ak`; `params_cost` refreshed for the 6-field datum; `plutus.json` rebuilt.
- **P6** — docs: `02`/`03`/`08`/`09` (final surface), migration `§17` in `cip113-api-changes-post-audit.md`, `CONTRACT_SURFACE_CHANGES.md` row.
- **`5c111f9`** — `ProgrammableLogicGlobalParams` reordered by read-frequency + renamed: `{ registry_node_cs(0), prog_logic_cred(1), transfer_cred(2), third_party_cred(3), unfracking_cred(4), upgrade_cred(5) }`.
- **Three-way dispatch (latest commit)** — `programmable_logic_global` → **`transfer`** (blueprint title `transfer.transfer.withdraw`; `ProgrammableLogicGlobalRedeemer` → single-constructor **`TransferRedeemer { params_idx, proofs }`**, bytes-compatible with the old `TransferAct`); `UnfrackingAct` removed — PLB gains **`SpendViaUnfracking`** (constructor 2, reads `unfracking_cred` field 4) so an unfracking tx is PLB + `unfracking` only; `SpendViaGlobal` → **`SpendViaTransfer`**. `issuance_mint` delegation re-pointed; all docs / comments / surface tracker updated.

## The two decisions worth close review

1. **`issuance_mint` delegation (P4, security-sensitive).** A mint can co-occur with a seize of the same policy; `validate_3rd_party`'s conservation check covers the minted subject tokens, so delegating custody to the `third_party` validator is sound — same "same registry-node index ⇒ same policy" reasoning as the old inline arm. Positive + negative tests isolate it by a single `registry_node_idx` delta (0 delegates / 1 falls back to local `no_escape`).
2. **No redeemer switch anywhere.** `TransferRedeemer` is a single constructor, so routing a seize or an unfracking to the transfer validator is unrepresentable at the type level; each path has its own validator and redeemer (`ThirdPartyRedeemer`, `UnfrackingRedeemer`). Pre-mint wire-format change: PLB's `BaseSpendRedeemer` has three constructors, and the PLG redeemer type is gone.

## Tests

Full suite green (**405/405 — 375 unit + 30 properties × 100 iterations, 0 warnings**). New/relocated: 26 third-party tests on the standalone validator; PLB dispatch + load-bearing-index + cross-arm rejection; `issuance_mint` third-party delegation (positive + negative); plus the review follow-ups below.

## Review follow-ups (2026-08-21, addressing @paolino's review)

- **Composed `issuance_mint × PLB × third_party` transaction** (`issuance_mint.test.ak`, `composed_seize_mint_*`): one coherent tx, every index (RefInput / `params_idx` / `registry_node_idx` / `wdrl_idx`) derived from the same layout, withdrawals in ledger order. Positive case keeps the mint at PLB; the one-token-escape mutation passes `issuance_mint` (it has delegated) and PLB (it has dispatched) and is rejected by `third_party` alone. Both also as 100-iteration properties over random mint/burn quantities, reference-input order and withdrawal position. The isolated delegation tests' `params_idx` now matches the coordination input's position.
- **`list.expect_at` is O(`wdrl_idx`)** — claims corrected in PLB, `types.ak`, `benchmarks.ak`, docs 02 / api-changes §16–17, `CONTRACT_SURFACE_CHANGES.md`; new width/position cost record `wdrl_idx_cost.test.ak` with a local reproduction of the pre-split scan as the comparison column.
- **Pairwise-distinct delegate credentials (`transfer_cred`, `third_party_cred`, `unfracking_cred`) is an assumption, not an invariant** — decided NOT to enforce on-chain (deployment / upgrade-authority responsibility); documented in `types.ak`, PLB, docs 02 / 08 / api-changes, and pinned by `plb_equal_delegate_creds_collapse_the_arms`.
- **Doc 09 `wdrl_idx` derivation** rewritten: ledger order is script credentials before key credentials, bytewise within each, over the complete withdrawal set — with a correct helper; the bech32 `.sort()` example is gone. Shared comparator for tests in `fixture.ak`.
- **PLB properties** (`programmable_logic_base.test.ak`): random widths (1–12) with mixed Script/VKey neighbours, random params-ref position, all three arms — the recomputed index accepts, every other in-range index rejects; out-of-range / negative index fail tests; cross-arm negatives between all pairs.
- **Golden raw-`Data` layout tests** (`layout.test.ak`): six params fields incl. `third_party_cred` / `upgrade_cred`, `BaseSpendRedeemer` constructors 0/1/2, `TransferRedeemer` (single constructor), `ThirdPartyRedeemer`, `UnfrackingRedeemer`, `MintingRegistryProof` — encode and decode directions.
- `CONTRACT_SURFACE_CHANGES.md` consolidated tables + SDK map refreshed to this head (PLB typed redeemer, `UnfrackingAct` = 1, 6-field params, seizure via `third_party`; #109 section added); dead `upgrade_cred_field` accessor removed.

Still out of scope here (tracked separately): Lean/FV update + `verify` in CI on this stack (note for the theorem names: the arms are now `SpendViaTransfer` / `SpendViaThirdParty` / `SpendViaUnfracking` and the transfer validator's title is `transfer.transfer.withdraw`); porting #101's ThirdPartyAct fuzz properties to `third_party.test.ak`.

## Out of scope (tracked separately)

- **FV / Lean** — the containment proofs rest on the `PLB → PLG` invariant this deliberately breaks; needs updating for PLB-dispatch + standalone `third_party`.
- **SDK / off-chain** — `cip113-sdk-ts` + Java backend need the new PLB sum-type redeemer, the `third_party` deploy + reference script, and `params_idx`/`wdrl_idx` derivation.


paolino (changes requested):
The split looks internally coherent, but I don't think this should merge as formally verified yet. The security-sensitive composition and verification story are incomplete.

## Blocking findings

1. **PR #110 invalidates the existing Lean forwarding theorem.**

   PR #101 proves that PLB acceptance forces the PLG withdrawal. That statement becomes false once PLB may dispatch directly to `third_party`. This PR contains no Lean sources, and PR #101 still imports the old parameterised, redeemer-independent PLB and old PLG bytecode.

   The Lean update can live in #101 or a linked follow-up, but it should be a required dependency before the new architecture is described as formally verified.

2. **The issuance delegation test is not a composed validation test.**

   In `validators/issuance_mint.test.ak:693`, `refinput_mint_delegates_to_third_party_seize` invokes only `issuance_mint`, deliberately sends the minted tokens outside PLB, and relies on the statement that `third_party` would catch the escape in a real transaction. Its `ThirdPartyRedeemer.params_idx` is also `0`, while the coordination input is at index `1`, so executing `third_party` on this fixture would reject before reaching containment.

   Please add a transaction that executes `issuance_mint + PLB + third_party` with coherent indices. The positive case must keep the mint at PLB; a one-token escape mutation must fail specifically in `third_party`.

## Other findings

- `documentation/09-DEVELOPING-SUBSTANDARDS.md:832` computes `wdrl_idx` by applying JavaScript `.sort()` to two Bech32 reward-address strings. The ledger orders the complete withdrawal map by `Credential`: Script credentials first, then VerificationKey credentials, hashes bytewise within each constructor. Extra withdrawals also shift the index. The example will produce invalid PLB redeemers for legitimate transactions.

- `list.expect_at` is not O(1), despite the comments in `validators/programmable_logic_base.ak` and the PR description. Stdlib recursively removes list cells until reaching the index, so it is O(`wdrl_idx`). The composition benchmarks hard-code `wdrl_idx = 0`, and therefore do not measure the claimed off-first-position saving. Please describe it as a direct indexed lookup with fewer comparisons and add a width/target-position benchmark.

- The claim that a wrong arm necessarily rejects assumes `transfer_cred != third_party_cred`. Neither `protocol_params_mint` nor `coordination_spend` enforces that. If the fields are equal, changing the arm still passes PLB. Either enforce distinct credentials or qualify the documentation, fuzz properties, and Lean theorem with that assumption.

- `CONTRACT_SURFACE_CHANGES.md:192-244` remains stale after the newly added PLG-split section: it still describes PLB as untyped, `UnfrackingAct` as constructor 2, the old three-field params datum, and seizure through PLG. The PR body also calls `third_party_cred` field 5; after the final reorder it is field 3.

## Lean coverage requested

Over freshly extracted compiled artifacts, I would expect:

- `plb_global_dispatch_sound`: acceptance under `SpendViaGlobal {p,w}` implies authenticated params at `p` and `withdrawals[w] = transfer_cred`.
- `plb_third_party_dispatch_sound`: the analogous theorem for `third_party_cred`.
- Dispatcher polarity controls for wrong, negative and out-of-range indices, stale credentials, and cross-arm credentials.
- Refreshed PLG transfer escape/conservation and two-owner authorisation theorems using the six-field datum and new redeemer encoding.
- Standalone third-party containment: subject-policy input plus mint/burn remains in PLB outputs; non-subject policies, address, datum and reference script are preserved; paired-output lovelace cannot decrease.
- `issuance_third_party_delegation_sound`: same-node delegation authenticates the same policy and the standalone validator covers its mint custody.
- Golden decode checks for both `BaseSpendRedeemer` constructors, `ThirdPartyRedeemer`, `UnfrackingAct` constructor 1, and the six-field params layout.
- Updated flats/manifest closure and mutants that remove the indexed credential check, conservation check, or registry-node equality.

## Aiken fuzz/property coverage requested

PR #101 already has useful ThirdPartyAct properties that can be ported to `third_party.test.ak`: arbitrary Ada top-up/drain, arbitrary seizure split, arbitrary escaped quantity, and arbitrary co-resident-token preservation/delta.

Please also add:

- Random params-reference and withdrawal positions for both PLB arms: the recomputed index accepts and every other index rejects.
- Mixed Script/VKey withdrawal lists using the actual credential comparator.
- Equal delegate credentials, to force a decision on whether distinctness is required.
- Golden raw-`Data` layout tests for all six params fields and redeemer constructor/field order. The current params-cost tests omit direct checks for `third_party_cred` and `upgrade_cred`.
- Full `issuance_mint × PLB × third_party` composition with randomized mint/burn quantities and index placement, plus a one-token escape mutation.

The current CI job at `5c111f9` is green (364/364 scenarios in both environments), but `programmable_logic_base.test`, `third_party.test`, and `issuance_mint.test` each report zero property tests.

## CI integration requirement

The Lean verification workflow must be part of normal validator CI, not isolated on PR #101. PR #110 currently has only the Aiken `build` check because its base branch, `feat/upgradability-in-place`, does not contain `.github/workflows/formal-verification.yml` or the formal-verification tree. The existing workflow already watches `validators/**`, `lib/**`, and `plutus.json`, so it would run for these changes if the branch inherited it.

Suggested integration path: land #101, rebase `feat/upgradability-in-place` onto the updated `main`, then rebase this branch onto that refreshed base. Alternatively, rebase the formal-verification branch onto this stack and bring its workflow and verification tree here. In either case, refresh the flats/manifest and update the Lean statements until `verify` is green on this exact validator head. Finally, make the `verify` job a required branch-protection check so future validator-byte changes cannot pass CI without re-establishing the Lean claims.

validators/issuance_mint.ak
```diff
@@ -192,7 +192,7 @@ fn verify_registry_node(
 /// at runtime and issuance_mint can safely delegate.
 ///
 /// PLGlobal's credential is NOT a parameter: it is the LIVE one read from
-/// the coordination UTxO's datum (field 3, `prog_logic_global_cred`), so
+/// the coordination UTxO's datum (field 2, `transfer_cred`), so
```
Kammerlo: Same. Is this correct?

validators/issuance_mint.ak
```diff
@@ -27,7 +27,7 @@ validator issuance_mint(
   // the same parameter PLB and PLGlobal take. Read ONLY inside the
   // `RefInput` delegation branch, where it locates the coordination UTxO
   // among `reference_inputs` and reads the LIVE PLGlobal credential from
-  // its datum (field 3, `prog_logic_global_cred`). Delegation then keys on
+  // its datum (field 2, `transfer_cred`). Delegation then keys on
```
Kammerlo: Is this comment correct? Shouldn't this be field 1? Prog_logic_cred? Same a little bit downwards

validators/coordination_spend.ak
```diff
@@ -96,22 +96,27 @@ validator coordination_spend(_nonce: ByteArray) {
...
       // never appear in tx.withdrawals. Writing one here is a ONE-WAY BRICK:
-      // an unsatisfiable upgrade_logic_cred makes this validator's own
+      // an unsatisfiable upgrade_cred makes this validator's own
       // authority check below permanently unsatisfiable, with no repair path,
-      // and an unsatisfiable prog_logic_global_cred makes every programmable
+      // and an unsatisfiable transfer_cred makes every programmable
       // token UTxO unspendable. Same rule linked_list.ak already applies to
       // every registry-node credential.
-      is_28_byte_credential(new_params.prog_logic_global_cred)?,
+      is_28_byte_credential(new_params.transfer_cred)?,
```
Kammerlo, line 104: I think it makes sense to check if the new parameters are actually scripts. Like 
```
expect Script(_) = new_params.transfer_cred
expect Script(_) = new_params.third_party_cred
expect Script(_) = new_params.unfracking_cred
```

nemo83: thanks for this, I just had tiem to review. in minting we're not doing it. so for the moment, I might leave it as is (this is testable in preview etc.)

Kammerlo (approved):

PR #111, opened 2026-08-20
fix: constrain PLB output shape at creation (issue #106)

### LINKED ISSUE #106: Seizure (ThirdPartyAct) can be rendered unsatisfiable by holder-controlled UTxO shape
### Summary

A token holder can shape their own programmable UTxOs so that an issuer's seizure of them becomes impossible to construct, **while the holder keeps using those UTxOs normally**. Because seizure is what makes this standard usable for regulated/compliance assets, a cheap, repeatable, preemptive way to defeat it undermines the core guarantee: a non-compliant or sanctioned holder could make their assets unseizable yet still spendable by themselves.

The load-bearing property is *asymmetry*: the vector is real where the holder can spend the UTxO but the issuer cannot seize it. That holds for the **datum** and the **reference script**. It does **not** cleanly hold for tokens (see below).

### Mechanism

- **Creation (`TransferAct`, holder-witnessed)** imposes no constraint on the shape of the PLB outputs it produces. `collect_output_assets` (`lib/assets.ak`) accepts an output on payment credential + inline stake credential alone — datum kind, datum size, reference script, and token set are all unconstrained.
- **Seizure (`ThirdPartyAct`)** must spend each targeted input and produce a paired continuing output that reproduces it byte-for-byte (`third_party.ak:163-165`: `output.address == input.address`, `output.datum == input.datum`, `output.reference_script == input.reference_script`, plus conservation of every non-seized policy). `unfracking.ak:265-267` imposes the same triple.

Pairing is atomic per-UTxO, so the issuer cannot dodge the cost by batching differently.

### Vectors

**1. Datum hash — real, asymmetric, ABSOLUTE, and the cheapest. (Highest priority.)**
Nothing constrains the *kind* of datum on a PLB output, so a holder can create their UTxO with `datum = DatumHash(hash(secret))` for a secret they keep (creating a datum-hash output does not reveal the preimage). To *spend* a datum-hash UTxO, the **ledger** (Alonzo+) requires the matching datum preimage in the transaction witness set — enforced at phase 1, before any validator runs. The holder knows the secret and can spend via `TransferAct`; the issuer does not and therefore **cannot form the seizure transaction at all**. This is not a size/cost issue — it is an unconditional block, costs a 32-byte hash, and is repeatable (rotate to a fresh secret on each self-spend, so live UTxOs are always unseizable).
*Status in code: **not prevented.** `grep` finds no `DatumHash` handling anywhere in `lib/` or `validators/`.*

**2. Reference script — real, asymmetric (size).** The holder attaches a large `reference_script`. Seizure must embed the full script bytes in the continuing output → over `maxTxSize`. `TransferAct` need not reproduce it, so the holder keeps spending while the issuer cannot seize.

**3. Large inline datum — real, asymmetric (size).** The holder attaches a large inline datum. Seizure must reproduce it → over `maxTxSize`. Softer than vector 1 (bounded by tx-size rather than absolute), and only relevant once datum hashes are banned.

**4. Tokens — could bite at the margin, but self-limiting; not a clean evasion.** Loading many non-programmable tokens raises the seizure conservation-walk cost. But it is bounded and symmetric:
- **Symmetric burden.** `verify_proofs` in the `TransferAct` path requires one registry proof per distinct input policy (a non-membership proof for each junk policy) plus a per-policy token comparison — so stuffing heavy enough to block `ThirdPartyAct` also blocks the holder's own `TransferAct`. The UTxO self-freezes rather than becoming holder-usable / issuer-unseizable.
- **`maxValueSize`** caps a single UTxO's value (~5 KB), bounding policies/tokens so the walk stays within the per-transaction execution budget; the issuer seizes one UTxO at a time.

The worst realistic outcome is a UTxO nobody can spend — which, for a confiscation target, is an acceptable (frozen) end state, not an escape. `UnfrackingAct` is the intended recovery for legitimately fractured UTxOs but is holder-driven; that does not change the conclusion, since the hostile outcome is self-freeze. **Monitor; not a base-layer must-fix.**

### Impact

Vectors 1–3 defeat the issuer's seizure capability while the holder retains use of the asset. Vector 1 is the most severe: unconditional, ~free, and currently unprevented. The freeze fallback (transfer logic declining) is strictly weaker — it immobilizes but does not let the issuer recover/confiscate. For a compliance-oriented deployment this is a critical gap.

### Affected code

- `lib/assets.ak` — `collect_output_assets` (creation-side; datum kind/size + reference script unconstrained).
- `validators/programmable_logic/third_party.ak:163-165` — seizure reproduction.
- `validators/programmable_logic/unfracking.ak:265-267` — same reproduction on the unfracking path.
- `validators/issuance_mint.ak` — `no_escape` (mint-side output creation, same lack of shape constraint).

### Proposed direction — constrain at creation (NOT relax at seizure)

The seizure reproduction check must stay strict: allowing seizure to drop or alter the datum/reference script would let any programmable-token admin tamper with a **co-resident** policy's datum in a multi-asset UTxO (e.g. a scam token registered solely to "seize" itself and wipe a victim's datum for a legitimate co-resident asset). So the fix constrains what can be created, not what seizure must reproduce.

1. **Datum kind — require `NoDatum` or `InlineDatum(_)`; reject `DatumHash(_)`** at every PLB-output creation site (`collect_output_assets`, `issuance_mint`'s `no_escape`, unfracking destination). No legitimate programmable-token use needs a datum hash (inline is strictly better: readable, no preimage custody). Closes vector 1 with no downside. **Highest priority.**
2. **Reference script — ban on PLB outputs at creation** (`reference_script == None`, same sites). Token UTxOs have no legitimate need for a reference script; makes the seizure reproduction trivially `None == None`. Closes vector 2.
3. **Inline datum size — bound the serialized length** (`serialise_data(datum)` length ≤ N), keeping byte-for-byte reproduction at seizure. N large enough for legitimate per-UTxO substandard state, small enough that reproduction is always affordable. The exact bound (and whether PLB UTxOs should carry inline datums at all vs. a smaller state mechanism) is an open design decision. Closes vector 3.
4. **Tokens — no base-layer change** (see vector 4); revisit only if a concrete holder-usable / issuer-unseizable construction survives the symmetric-burden argument.

The standard is not yet deployed, so creation constraints can be enforced from genesis with no migration.

MavenRain:
Fixes #106.

## Summary

A holder can shape their own PLB UTxOs (datum hash, reference script) so that issuer seizure (`ThirdPartyAct`) becomes impossible to construct while the holder keeps spending via `TransferAct`. This PR closes vectors 1 and 2 of #106 the way the issue proposes: constrain the output shape at every PLB-output **creation** site, and leave the seizure/unfracking byte-for-byte **reproduction** checks strict (relaxing them would let a co-resident policy's admin tamper with a victim's datum).

## Fix

New total predicate `utils.is_seizable_shape(output)`:

- datum is `NoDatum` or `InlineDatum(_)` — never `DatumHash(_)`. Spending a datum-hash output requires the preimage in the witness set (ledger phase 1), which only the holder has, so the issuer cannot even form the seizure transaction.
- `reference_script == None`. Seizure must reproduce the full script bytes on the paired continuing output, so a large script pushes every seizure over `maxTxSize` while `TransferAct` spends unaffected.

Enforced at the three creation sites named in the issue:

1. `lib/assets.ak` `collect_output_assets` — the `TransferAct` output scan.
2. `validators/issuance_mint.ak` `no_escape` — mint-side PLB outputs.
3. `validators/programmable_logic/unfracking.ak` — the destination fold (unpaired outputs at the owner address). The paired continuing outputs reproduce their inputs byte-for-byte, so they inherit the shape by induction.

The standard is not deployed, so this is genesis enforcement with no migration.

## Not in this PR (follow-ups)

- **Vector 3 (inline-datum size bound):** the issue flags the bound `N` as an open design decision (and whether PLB UTxOs should carry inline datums at all). Deliberately left out; happy to follow up once the bound is decided.
- **`ThirdPartyAct` destination outputs:** the unpaired outputs receiving seized tokens are created by the seizing issuer, so a bad shape there only harms the actor that chose it; the issue scopes them out. Worth a follow-up decision if you want the "no PLB UTxO ever has an unseizable shape" invariant to be unconditional.
- `plutus.json` is left to the usual `chore: latest plutus json` regeneration.

## Testing

Red-first: the six repro tests were written and run **before** the fix — all six failed (the validators accept holder-shaped outputs today), with a green shape control on the happy path. After the fix the full suite is green: **369 passed, 0 failed** (`aiken check`, v1.1.23), `aiken fmt --check` clean.

- `transfer_act_fails_datum_hash_output` / `transfer_act_fails_reference_script_output` (+ `transfer_act_shape_control_succeeds` control) — end-to-end `TransferAct` through `programmable_logic_global.withdraw`.
- `mint_output_datum_hash_fails` / `mint_output_reference_script_fails` — `no_escape` path.
- `unfracking_fails_destination_datum_hash` / `unfracking_fails_destination_reference_script` — unfracking destination, one delta per test off the shared fixture.
- Four unit tests pin `is_seizable_shape` itself (both accepts, both rejects).

### DIFF
--- a/lib/assets.ak
+++ b/lib/assets.ak
@@ -6,6 +6,7 @@ use cardano/transaction.{Output}
 use env
 use list
 use tokens.{Tokens}
+use utils
 
 pub type Assets =
   Pairs<PolicyId, Tokens>
@@ -184,6 +185,13 @@ pub fn has_policy_nft_strictly(
   actual_asset_name == asset_name
 }
 
+/// Collect the assets of every output at the programmable-logic base
+/// credential. Every selected output must carry an inline stake credential
+/// (so ownership is well-defined) and a seizable shape (issue #106): datum
+/// absent or inline — never a hash — and no reference script. Constraining
+/// the shape HERE, at the creation site, is what keeps `ThirdPartyAct`'s
+/// byte-for-byte reproduction satisfiable for the issuer; see
+/// `utils.is_seizable_shape`.
 pub fn collect_output_assets(
   outputs: List<Output>,
   prog_logic_cred: Credential,
@@ -194,6 +202,7 @@ pub fn collect_output_assets(
     fn(output, select, discard) {
       if output.address.payment_credential == prog_logic_cred {
         expect Some(Inline(..)) = output.address.stake_credential
+        expect utils.is_seizable_shape(output)?
         select(output)
       } else {
         discard()
--- a/lib/utils.ak
+++ b/lib/utils.ak
@@ -5,7 +5,35 @@ use aiken/builtin.{tail_list}
 use aiken/collection/dict
 use cardano/address.{Credential, Script}
 use cardano/assets.{PolicyId, Value}
-use cardano/transaction.{InlineDatum, Output}
+use cardano/transaction.{DatumHash, InlineDatum, NoDatum, Output}
+
+/// A programmable-token (PLB) output has a *seizable shape* iff its datum
+/// is absent or inline — never a hash — and it carries no reference script.
+///
+/// Both constraints exist for the same reason (issue #106): seizure
+/// (`ThirdPartyAct`) and unfracking must reproduce `address`, `datum` and
+/// `reference_script` byte-for-byte on the paired continuing output, so any
+/// shape only the holder can reproduce (or afford) defeats them while
+/// `TransferAct` keeps working:
+///
+///   * `DatumHash`: the ledger (phase 1) demands the datum preimage in the
+///     witness set of any transaction SPENDING the output. The holder keeps
+///     the secret, so the issuer cannot even form the seizure transaction.
+///   * reference script: the continuing output must embed the full script
+///     bytes, so a large script pushes every seizure over `maxTxSize`.
+///
+/// Enforced at every PLB-output creation site (`assets.collect_output_assets`,
+/// `issuance_mint`'s `no_escape`, the unfracking destination); the
+/// reproduction sites then preserve it by induction, so no live PLB UTxO can
+/// ever take an unseizable shape.
+pub fn is_seizable_shape(output: Output) -> Bool {
+  let datum_admissible = when output.datum is {
+    NoDatum -> True
+    DatumHash(_) -> False
+    InlineDatum(_) -> True
+  }
+  datum_admissible && output.reference_script == None
+}
 
 /// Get the datum from an output, expecting an inline datum
 pub fn expect_inline_datum(output: Output) -> Data {
--- a/lib/utils.test.ak
+++ b/lib/utils.test.ak
@@ -2,11 +2,60 @@
 
 use aiken/collection/dict
 use aiken/collection/list
+use cardano/address.{Address, Script}
 use cardano/assets.{
   PolicyId, Value, ada_asset_name, ada_policy_id, add, from_asset, zero,
 }
+use cardano/transaction.{DatumHash, InlineDatum, NoDatum, Output}
 use utils.{
-  bytearray_lt, count_unique_tokens, has_currency_symbol, mint_has_policy,
+  bytearray_lt, count_unique_tokens, has_currency_symbol, is_seizable_shape,
+  mint_has_policy,
+}
+
+/// A well-shaped PLB-like output: no datum, no reference script.
+const some_plb_shaped_output: Output =
+  Output {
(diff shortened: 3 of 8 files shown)


MavenRain:
@nemo83 Wondering why this was closed w/o comment

Issue #108, opened 2026-08-19
Consider reserving a forward-compatible extension: Data field on the protocol datums before mint


nemo83:
### Context

The protocol's on-chain datums have a fixed field count, enforced by exact-arity casts (`expect _: ProgrammableLogicGlobalParams = datum`, `expect _: RegistryNode = datum`):

- `ProgrammableLogicGlobalParams` — `registry_node_cs`, `prog_logic_cred`, `unfracking_cred`, `prog_logic_global_cred`, `upgrade_logic_cred`
- `RegistryNode` — `key`, `next`, `minting_logic_script`, `transfer_logic_script`, `third_party_transfer_logic_script`, `unfracking_logic_script`, `global_state_cs`

These shapes are pinned to token addresses through a chain of hashes: the coordination-spend hash fixes the coordination address, which parameterizes `protocol_params_mint`, which fixes the params-NFT policy, which parameterizes `programmable_logic_base`, whose hash *is* every programmable-token address. So changing either datum's shape changes the PLB hash and therefore every token address.

### The one-way door

Before the one-shot params NFT is minted, fields can be appended freely (this is how the params datum grew from 3 to 5 fields during the upgradability work). **After mint, the schema is frozen for the deployment's lifetime** — adding a field would require redeploying the whole protocol and migrating every token to new addresses.

The in-place upgrade mechanism (`coordination_spend`) covers *logic* evolution (PLG and the credentials can be swapped), but not *schema* evolution.

### Proposal

Consider appending a trailing `extension: Data` field to one or both records before mint:

```aiken
pub type ProgrammableLogicGlobalParams {
  // ...existing fields...
  extension: Data,   // reserved, opaque, empty at bootstrap
}
```

Because it is typed `Data`, current validators require it to be present but accept any content, so they carry it without interpreting it. A future PLG (swappable via the upgrade mechanism, since PLB reads the PLG credential from the datum) could later read additive config out of it — without any hash changing.

**Reach:** only config that a *swappable* validator (PLG) reads is reachable this way; the frozen validators (`registry_mint`/`registry_spend`, `coordination_spend`, PLB) can never interpret new meaning from it.

**Cost:** a permanently-present, always-open opaque field on every datum from day one (it must be left unconstrained to remain fillable later).

### Decision

[...]

PR #112, opened 2026-08-21
Reserve a trailing extension: Data field on the protocol datums (#108)

MavenRain:
### What

Implements the proposal in #108: a trailing, reserved `extension: Data` field on both protocol datums, appended before the one-shot params mint makes the schema permanent.

- `ProgrammableLogicGlobalParams` gains field index 5, `extension: Data`.
- `RegistryNode` gains field index 7, `extension: Data`.
- `utils.empty_extension` (`Void`, encoding `Constr 0 []`) is the single spelling of "empty" and the bootstrap value everywhere.

### Why now

The issue describes the one-way door: after mint, adding a field means redeploying the protocol and migrating every token, because both shapes are pinned to every token address through the PLB hash chain.  The in-place upgrade mechanism covers logic evolution only.  This field is the schema affordance, and it only works if it ships before mint.

### Behaviour in the frozen validators

- They carry the field but never read it.  The positional accessors (`with_key_and_*` in `registry_node.ak`, `*_field` in `params.ak`) read from the front of the field list, so no index moves.
- `coordination_spend`: the exact-arity cast requires the field on the continuing datum, and the frozen-anchor checks do not cover it, so the sitting upgrade authority may rewrite it.  That is what makes it fillable after mint.  No code change beyond the type.
- `registry_mint` insert: issuer-chosen and deliberately unconstrained.  A shape pinned by a frozen validator would defeat the field's purpose.
- Covering node during an insert: frozen by the existing spread equality in `is_updated_directory_node`.  No new code.
- Registry update path: mutable, like the other logic fields (`is_field_updated_registry_node` copies `new.extension`).
- Origin node: pinned to `empty_extension` by the existing single-equality origin check.  No new code.

### Testing

6 new tests;  full suite green (`aiken check` 0 failed, `aiken fmt` clean):

- `fails_with_old_five_field_datum_shape`: a bootstrap datum in the pre-change 5-field shape is rejected at mint.  This is the door closing.
- `coord_upgrade_can_fill_extension`: the upgrade path rewrites the field.
- `is_inserted_directory_node_accepts_nonempty_extension` and `is_field_updated_registry_node_allows_extension_rewrite`: both registry mutation paths accept content.
- `is_updated_directory_node_rejects_extension_change` and `is_origin_directory_node_rejects_nonempty_extension`: the frozen shapes still reject tampering.

Mechanical fallout: every full record literal in tests and fixtures gains `extension: empty_extension` (55 sites).  The `any_registry_node` fuzzer pins the extension to empty;  widening it to arbitrary `Data` is a possible follow-up.

### Open points for the decision

- Both records get the field here.  Trimming to one record is a small diff if the decision lands that way.
- `Void` as the empty sentinel is a choice, not a requirement.  Any stable `Data` term works.
- `plutus.json` regeneration is left to the maintainer chore flow, matching prior PRs.

Addresses #108.


MavenRain:
@nemo83 Wondering why this was closed w/o comment

PR #117, opened 2026-09-02
Thin the trust chain: PLG dispatcher, merged registry, four-field params datum

nemo83:
Everything frozen gets baked into the artifact that owns it; the protocol-params datum keeps only live wiring. Four changes, one hash cascade — they ship together because each one's saving depends on the others.

**Docs are deliberately untouched** beyond the identifier rename sweep. They land as a single pass once the code is reviewed.

## 1. PLB dispatches through a dispatcher

```
PLB    reads params[plg_cred, index 0] -> requires PLG's withdraw-0.  That is the whole validator.
PLG    transfer / third_party / unfracking hashes BAKED as compile-time parameters;
       each arm proves the right delegate ran. Reads no datum.
DELEG  prog_logic_cred / registry_node_cs / max_inline_datum_bytes BAKED. Read no datum.
```

`BaseSpendRedeemer` becomes a record. The three-constructor enum is gone, and with it the **pairwise-distinctness assumption** PLB's arms rested on — previously, two equal delegate credentials collapsed the arms and nothing on-chain enforced otherwise. That assumption is now designed out rather than merely untested.

transfer tx **−27 B, −7.80M cpu**; PLB **−220 B and −0.95M cpu per input**.

## 2. `coordination_spend` → `protocol_params_spend`

Mirrors the existing `registry_mint` / `registry_spend` pair. "Coordination UTxO" is retired from the code — it and the protocol-params UTxO were always one thing under two names.

## 3. `registry_mint` + `registry_spend` → one `registry`

`registry_mint` was parameterized by `registry_spend_cred`, so `registry_spend` could not take `registry_node_cs` as a parameter without closing a hash cycle. That cycle — not upgradability — was the only reason it read the params datum.

Merged, the two hashes are one and each handler names itself: the mint handler is handed `policy_id`, and the spend handler derives the policy from its own input's payment credential. That binding is total, not a check — the ledger only invokes the handler for UTxOs at this address.

| registry action | scripts loaded | before | after | delta |
|---|---|---:|---:|---:|
| Insert | mint + spend | 3481 B | 2674 B | **−807 B (−880 applied)** |
| Init | mint only | 1928 B | 2674 B | +746 B |
| Update | spend only | 1553 B | 2674 B | +1121 B |

Insert loads one reference script instead of two, with shared helpers deduplicated. Init and Update now carry the handler they do not use — the accepted price, agreed before the work started. Node spend **−33.5%** cpu, node update **−17.0%**, like for like.

The registry no longer touches the protocol-params chain at all, so it can be compiled right after `issuance_cbor_hex_mint`.

## 4. Params datum 6 → 4 fields

| idx | field | reader |
|---|---|---|
| 0 | `plg_cred` | `programmable_logic_base`, **per programmable input** |
| 1 | `transfer_cred` | `issuance_mint` |
| 2 | `third_party_cred` | `issuance_mint` |
| 3 | `upgrade_cred` | `protocol_params_spend` |

`prog_logic_cred` had no runtime reader at all; `registry_node_cs` lost its last one to the merge. Both were frozen by `protocol_params_spend`, so removing them converts **two rails into two unrepresentable states**. Every remaining field is mutable live wiring with a named reader.

`plg_cred` at index 0 saves PLB **227,326 cpu per programmable input** — the only per-input datum read in the protocol. One `tail_list` step over this datum measures 113,663 cpu, linear across depths 0–3.

Finding 04's precise delegation is untouched: `issuance_mint` still reads the two delegate credentials at runtime and still keys on the delegate's own redeemer.

## Two undefended guards, both pre-existing, now closed

**The params-NFT authentication in `with_protocol_params_fields`.** Neutering it left the entire suite green. A transaction with no params UTxO anywhere, carrying an attacker's four-field datum on an ordinary reference input, would have authorised spending any programmable UTxO on the deployment. The test that named the guard was green for the wrong reason: its fixture lacked *both* the NFT and an inline datum, so removing the guard merely moved the abort one line down — and a `fail` test in Aiken passes on an abort. Now killed by exactly the two tests that name it.

**`VerificationKey(_) -> False` in the registry spend handler** — the origin node's only lock against an unsigned rewrite of its own fields. No test changed state when it was removed.

## Review notes

- **Two off-chain breaks are SILENT.** `SpendViaTransfer { params_idx, wdrl_idx }` and the new `BaseSpendRedeemer { params_idx, wdrl_idx }` are both "constructor 0 with two ints", so a stale builder decodes fine and fails later as a withdrawal-index mismatch. And old datum indices 1, 2 and 3 all keep the `Credential` shape while changing meaning — index 2 is worst, since a builder computing PLB's dispatch target silently receives `third_party_cred`. Both are tabulated in `CONTRACT_SURFACE_CHANGES.md`.
- **The registry policy id and the registry node address are now one value.** An SDK deriving them from two applied hashes must collapse to one derivation.
- **Genesis-only.** Every hash moves; substandards recompile with new constants. No substandard reads the params datum today — all seven across the platform and CMTA repos take what they need as compile-time parameters — so this withdraws an option nobody took.
- **The wire field order is defended by exactly two tests.** A self-consistent reorder of the record is invisible to 483 of 485 tests, because every on-chain reader goes through the accessors and agrees with itself. `layout.test.ak` carries a do-not-trim note.

## Verification

`aiken check -D` and `aiken check -D --env with_assertions`: **485/485, zero failures**. `aiken fmt --check` clean. `plutus.json` regenerated and reproducible from source. The committed tree was checked out in isolation and built green independently of the working tree.

Both slices went through a worker/auditor pair; each bounced once and neither finding recurred. Guards were verified by mutation rather than inspection — 16 mutants on the registry merge, 11 on the datum reduction.

https://claude.ai/code/session_017k2jCrHcccLJJ4fQBYHzvD


validators/programmable_logic_global.ak
```diff
@@ -0,0 +1,82 @@
...
+    // One lookup, against a hash that is part of this script's identity. A
+    // redeemer naming the wrong action resolves to a delegate that is not
+    // present, so it can only invalidate its own transaction.
+    let required = when redeemer is {
+      TransferAct -> transfer_hash
+      ThirdPartyAct -> third_party_hash
+      UnfrackingAct -> unfracking_hash
+    }
+
+    self.withdrawals |> pairs.has_key_or_fail(Script(required))
```
nemo83, line 69: There should always be at least 2 withdraw-0 scripts. So wondering if we should introduce an index here....

Kammerlo (approved):

validators/registry.ak
```diff
@@ -0,0 +1,246 @@
...
+validator registry(utxo_ref: OutputReference, issuance_cbor_hex_cs: PolicyId) {
+  mint(redeemer: RegistryRedeemer, policy_id: PolicyId, self: Transaction) {
+    trace @"Starting registry mint validation"
+    when redeemer is {
+      RegistryInit -> {
+        // Ensure this is a one-shot minting policy by checking that utxo_ref is spent
+        let is_utxo_consumed = list.any(
+          self.inputs,
+          fn(input) { input.output_reference == utxo_ref },
+        )
```
nemo83: can it be replace with list.expect_any?

validators/registry.ak
```diff
@@ -0,0 +1,246 @@
...
+          )?,
+        }
+      }
+      RegistryInsert { key, minting_logic_script } -> {
+        // Find the issuance CBOR hex reference input
+        let issuance_ref_input = list.expect_find(
+          self.reference_inputs,
+          fn(input) {
+            let tokens = assets.tokens(input.output.value, issuance_cbor_hex_cs)
+            !dict.is_empty(tokens)
```
nemo83: is this the most efficient way of doing it?

validators/registry.ak
```diff
@@ -0,0 +1,246 @@
...
+        // Whether a first mint of `key` occurs in the same tx is NOT
+        // checked here: if it does, issuance_mint validates it as usual.
+        expect
+          list.any(
+            self.withdrawals,
+            fn(w) {
+              let Pair(cred, _amount) = w
+              cred == minting_logic_script
+            },
+          )
```
nemo83: we do the same (or similar in other places) is this the most efficient? is there a reusable function?

validators/registry.ak
```diff
@@ -0,0 +1,246 @@
...
+
+    when
+      assets.to_dict(self.mint)
+        |> dict.get(registry_node_cs)
+    is {
+      // A registry node NFT is minted → we are the covering-node spend of an
+      // insert; the `mint` handler validates the linked-list structure.
+      Some(minted_assets) -> {
+        expect [Pair(_, amount)] = dict.to_pairs(minted_assets)
+        amount > 0
```
nemo83, line 211: should amount be replaced with 1 => `expect [Pair (_, 1)]` = ... 
and then return `True` ?

Overkill?

nemo83 (commented):
Added some self-comments to be addressed

Kammerlo (approved):

Issue #131, opened 2026-09-07
Documentation sweep for the #117 and #118 validator-set changes


nemo83:
## What is stale

Two merged changes moved the validator set, and the narrative documentation has not caught up.

**From #117** (dispatcher, registry merge, four-field params datum):
- `registry_mint` and `registry_spend` no longer exist — they are one `registry` validator with `mint` and `spend` handlers.
- The protocol-params datum is four fields: `plg_cred`, `transfer_cred`, `third_party_cred`, `upgrade_cred`. `registry_node_cs` and `prog_logic_cred` are gone.
- `programmable_logic_global` is back, as a dispatcher parameterised by the three delegate hashes.

**From #118** (params pair merge):
- `protocol_params_mint` and `protocol_params_spend` no longer exist — they are one `protocol_params` validator.
- `params_spend_addr_hash` and `_nonce` are gone as parameters.
- The params NFT policy id and the params address's payment credential are the same value.

Known affected: `documentation/02-ARCHITECTURE.md`, `documentation/03-*`, `documentation/08-*`, `documentation/09-DEVELOPING-SUBSTANDARDS.md`, `documentation/design/upgradability-in-place.md`, `documentation/design/finding-17-unfracking-w0-delegation.md`, `documentation/cip113-api-changes-post-audit.md`, `README.md`, and the substandard upgrade guide.

## Also in scope

- The derivation guidance an SDK follows: after both merges, the registry address/policy and the params address/policy each collapse to a **single** derivation off one applied script.
- The deployment ordering section — the registry no longer depends on the protocol-params chain, and the params mint no longer depends on the params spend address.
- `CLAUDE.md`'s command table still cites a stale test count.

Documentation teaches the final surface only; the history of removed intermediate shapes stays out.

Issue #126, opened 2026-09-07
upgrade_multisig: duplicate signers lower the effective threshold, and threshold is unbounded


nemo83:
## Two defects in the same three lines

`validators/upgrade_multisig.ak`:

```aiken
let approvals = list.count(signers, fn(s) { list.has(self.extra_signatories, s) })
(approvals >= threshold)?
```

**1. Duplicates count more than once.** `list.count` counts occurrences in `signers`, not distinct approvers. With `signers = [A, A, B]` and `threshold = 2`, **A alone satisfies the multisig.** A duplicated key silently halves the security property, and nothing in the code or the type says so.

**2. `threshold` is unbounded.** `threshold = 0` (or negative) authorises every upgrade with zero signatures. `threshold > length(signers)` can never be satisfied — and since this script is the upgrade authority, that is a permanent brick with no repair path.

Both values are compile-time parameters, so neither is exploitable by a third party: they are chosen at deployment. They are still representable, unguarded, and invisible in the applied script.

## Proposed direction

Guard both in the withdraw handler — paid once per upgrade, on the rarest path in the protocol:

- signers must be distinct (compare `list.unique(signers)` length against `list.length(signers)`);
- `1 <= threshold <= length(signers)`.

Independent of whether the signer set later moves into a datum: the guards are worth having under either design.

## Acceptance

- Negative: a duplicated signer set is rejected even when the raw count meets the threshold.
- Negative: `threshold = 0` and `threshold > n` are both rejected.
- Positive: an ordinary M-of-N approval still passes, and M-1 signatures still fail.

Issue #124, opened 2026-09-07
Fixed three-action set limits protocol upgradability


nemo83:
The protocol represented programmable actions as a fixed three-action set spread across several places at once: one credential field per action in the protocol parameters, three constructors in the `programmable_logic_base` redeemer, and an exhaustive mapping between them in PLB itself.

An update could replace the validator behind an existing action, but could not add one. Adding an action meant changing the parameter schema, the PLB redeemer and the PLB dispatcher — and since PLB's credential is what every token address is built from, that would have required migrating every existing token UTxO.

nemo83:
Addressed in #117, along the second of the two lines proposed for it.

`programmable_logic_global` is back as the sole delegate PLB requires, parameterised with the transfer, third-party and unfracking script credentials. PLB now reads a single mutable `plg_cred` and its redeemer is a plain record (`{params_idx, wdrl_idx}`) with no action constructors. An upgrade can therefore deploy a dispatcher with a different action set and rewrite one datum field; PLB's hash, and every token address built from it, never moves.

One documented divergence: the recommendation had the three action fields collapsing into a single dispatcher credential. Two survive — `transfer_cred` and `third_party_cred` — because `issuance_mint` reads them at runtime for its precise-delegation check, which sits on a different trust chain and must answer "did a delegate cover *my* policy's registry node?". That dependency is the subject of a separate open ticket on issuance upgradability.

Issue #123, opened 2026-09-07
Redundant immutable fields in ProgrammableLogicGlobalParams


nemo83:
`ProgrammableLogicGlobalParams` carried two fields that could never change: `registry_node_cs` and `prog_logic_cred`. Both were read at runtime out of the protocol-params reference-input datum despite being fixed for the life of a deployment, which expanded the datum schema and obscured which parameters are genuinely upgradable.

nemo83:
Fixed in #117. Both fields are gone; the datum is now four fields — `plg_cred`, `transfer_cred`, `third_party_cred`, `upgrade_cred` — and each has exactly one named on-chain reader.

The values became compile-time parameters of the delegates that consume them, which is what the recommendation asked for. Two freeze rails in the params spend validator went with them: the forbidden states are now unrepresentable rather than rejected.

Note what this does and does not buy. The fields are unrepresentable, so no upgrade can edit them in place; the values remain reachable one hop away, because an upgrade that rewrites `plg_cred` installs a dispatch layer naming delegates applied with whatever values the authority chose. That is the same authority in the same transaction, not a separate escape.

PR #132, opened 2026-09-07
Two-phase authority handover: nominate, then the nominee activates itself (#125)

### LINKED ISSUE #125: protocol_params: authority handover is one-sided — require the incoming upgrade credential to co-sign
## Current behaviour

`protocol_params` (spend handler) authorises an upgrade by requiring the withdraw-0 of the credential named in the **current** datum:

```aiken
self.withdrawals |> pairs.has_key_or_fail(old_params.upgrade_cred)
```

That is correct for an ordinary parameter change. It is also the whole check when the upgrade **replaces `upgrade_cred` itself**. The incoming credential is validated for one thing only: that it is a well-formed 28-byte hash.

Covered today by `params_spend_authority_handover_succeeds`, which writes a new `upgrade_cred` with only the outgoing authority's withdrawal present.

## Why it matters

A handover to a credential that does not correspond to a deployed, runnable script hands the protocol to nobody. There is no repair path: the new `upgrade_cred` is the only credential that can authorise the next spend, so if it can never appear in `tx.withdrawals`, the protocol-params UTxO is frozen permanently. A typo, or a hash for a script that was never deployed, is enough.

The 28-byte rail catches malformed hashes; it cannot distinguish a deployed script from a plausible-looking hash.

## Proposed direction

When `new_params.upgrade_cred != old_params.upgrade_cred`, additionally require the **new** credential's withdraw-0 in the same transaction. That proves the incoming authority exists, is a script that can run, and consents to taking over.

- Preserves rotation entirely — only the handover transaction gains a requirement.
- Costs one comparison and one lookup, on the rarest path in the protocol.
- Ordinary upgrades are unaffected.

## Acceptance

- Positive: a handover carrying both withdrawals succeeds.
- Negative: a handover carrying only the outgoing authority's withdrawal fails.
- Negative: an ordinary upgrade still succeeds with only the sitting authority's withdrawal, proving the new requirement is scoped to the change.

nemo83:
Closes #125.

## The problem

`upgrade_cred` could be rewritten in one transaction, with nothing checked about the incoming credential but that it is 28 bytes long. A hash for a script nobody deployed — or a typo — becomes the only credential able to authorise the next spend of the protocol-params UTxO. There is no repair path.

## Why not the co-signature the issue proposes

The issue as filed asks for the incoming credential's withdraw-0 in the **same transaction** as the rewrite. That was rejected during design.

A co-signature can only be produced by a party that can be scheduled into someone else's transaction at a chosen moment. A **governance action** cannot, and neither can a multisig or cold quorum that assembles approval over days. So the rule would silently restrict who is permitted to become the upgrade authority — the opposite of what a mutable `upgrade_cred` exists for.

Audit-3 finding 05 asks for "evidence that the incoming authority is usable", without prescribing a mechanism. Two phases supply exactly that evidence, without the coordination requirement.

## What shipped

| phase | who authorises | what may change |
|---|---|---|
| **`NominateAuthority`** | sitting `upgrade_cred` | `pending_upgrade_cred` only — `Some(c)` nominates, `None` revokes |
| **`PromoteAuthority`** | the **nominee's own** withdraw-0 | `upgrade_cred` := nominee, nomination clears, nothing else |
| **`ProtocolUpgrade`** | sitting `upgrade_cred` | mutable wiring; never the authority, never the nomination |

The nominee activating itself is the evidence: it proves the credential exists, is a script that can run, and consents. Between the phases the sitting authority can revoke.

## Two things worth a reviewer's attention

**The datum field is appended, and that is load-bearing.** `pending_upgrade_cred: Option<Credential>` sits at index 4. `programmable_logic_base` and `issuance_mint` read the params datum through positional `head_list`/`tail_list` accessors over raw `unconstr_fields`, so a field added at the *end* costs them nothing, while one inserted anywhere else would silently deepen or re-point a hot accessor. Measured: `protocol_params` 1414 → 1921 B and **no other validator changed by a single byte**. Now stated as an append-only rule on the type and pinned by a golden test.

**The redeemer constrains, it does not merely describe.** The spend redeemer was ignored `Data`; it is now a field-less three-arm enum. Because `ProtocolUpgrade` freezes the nomination and `NominateAuthority` freezes everything else, a handover can never *begin* inside a transaction presenting itself as a routine parameter change — so the arm is a reliable public index of handover events, not a label.

This does **not** reopen #124 (fixed action set). That finding concerns the **PLB** redeemer, whose hash is baked into every token address, so its action set must stay open or an action change becomes a token-address migration. `protocol_params` *is* the upgrade path; replacing it is a redeploy by definition. Arms are field-less because every value is already in the continuing datum — repeating one would be the redundant-field mistake #123 removed.

## Behaviour changes (both deliberate)

- `params_spend_authority_handover_succeeds` **passed** and now fails, renamed `params_spend_fails_direct_authority_rewrite`. The one-step rewrite is gone.
- Nominating *and* changing a parameter in one transaction was allowed by an earlier draft of this branch and is refused now. Costs one extra transaction on the protocol's rarest path; buys that a handover always begins as its own visible transaction.

## Verification

512/512 in both environments (was 490 on `main`), `fmt --check` clean.

Every rail mutation-verified **against the full suite** — scoped runs undercount kills here:

| rail neutered | tests killed |
|---|---|
| genesis born at rest (`is_init`) | 1 |
| upgrade cannot move authority | 1 |
| upgrade cannot nominate | 1 |
| nomination freezes all else | 2 |
| promotion purity | 4 |
| nominee withdrawal required | 1 |
| sitting authority required | 2 |
| nominee 28-byte rail | 1 |

That process caught a defect in my own first draft: promotion purity written field-by-field killed **zero** mutants, because it was structurally unable to cover a field nobody had written yet. Both nomination and promotion are now single record equalities, the shape `linked_list.is_updated_directory_node` uses, so a sixth datum field is frozen automatically.

## Surface impact

`CONTRACT_SURFACE_CHANGES.md` carries the full table. For SDK consumers the sharp edge is the redeemer: a builder that passed `Void` now fails, and a *wrong* constructor index is worse than a missing one — it selects a different rule set rather than failing to decode. Indices are pinned by `layout_protocol_params_redeemer_constructors_are_field_less`.

Existing four-field params datums do not decode as the new type. That is redeploy-class, like every validator-byte change, and the validator's hash moved anyway.

## Not in this PR

- **Docs** — they still describe a six-field datum and `protocol_params_spend`, so they are behind #117 and #118 as well as this change. Folded into the doc sweep, #131, rather than half-fixed here.
- **`upgrade_multisig` guards** (#126) — independent of this design, so they land separately.

https://claude.ai/code/session_01DSRNmWRYUFaE9QaZzj5kFw


fabianbormann (approved):
It looks all good to me. A tiny question that came to my mind: With the changed flow, is is_28_byte_credential still needed, as there is now an actual signature check part of the flow at 2nd stage?

nemo83:
> It looks all good to me. A tiny question that came to my mind: With the changed flow, is is_28_byte_credential still needed, as there is now an actual signature check part of the flow at 2nd stage?

good point and it's true, the check is redundant. the same check (function) though is used at init time where the w0 hash isn't actually executed, because it's harmless, I think it's fine to keep it.

PR #133, opened 2026-09-08
Upgrade authority as a Sundae multisig tree in a config UTxO (#126, #127)

### LINKED ISSUE #126: upgrade_multisig: duplicate signers lower the effective threshold, and threshold is unbounded

### LINKED ISSUE #127: upgrade_multisig: move signer set and threshold out of compile-time parameters
## Context

`upgrade_multisig` takes `signers` and `threshold` as compile-time parameters, so any change to the signer set or the threshold produces a different applied script and therefore a different credential. Rotating a key is not a configuration change: it is a new deployment plus an authorised `upgrade_cred` handover.

That works — the authority is swappable, no token migration is involved — but it makes routine maintenance (adding a signer, retiring a compromised key, adjusting the threshold) an authority reinstantiation, with the latency and configuration risk that implies.

## Options to evaluate

**A. Datum-based configuration.** Keep a stable `upgrade_multisig` credential in `upgrade_cred`, parameterise it by an authenticated configuration NFT, and hold `{ signers, threshold }` in that UTxO's inline datum. Configuration updates authorise against the old datum and preserve the NFT, address and continuing output. Enforces valid key hashes, distinct signers, bounded configuration size, `1 <= threshold <= n`.

**B. Port native-script multisig semantics.** A structure in the shape of SundaeSwap's `multisig.ak` (`sundae/multisig`, the `satisfy` function) supports any combination rather than flat M-of-N, plus timelocks — so an upgrade could be approved now and executable only after a delay, mirroring how a hard fork triggers some time after approval. The logic can be copied rather than taken as a dependency.

**C. Keep the current reference authority** and document that signer/threshold changes require a new instance, with the guards from the sibling issue.

Note that under any option the core is unaffected: if the authority compiles to a withdraw-0 script, the protocol-params datum only ever references its hash. This is a replaceable component by design.

## Also worth deciding

Recovery when the threshold can no longer be met — a cold, timelocked, or governance branch. Neither the current design nor a self-authorised datum update provides one.

nemo83:
Closes #126. Closes #127. Also closes the unfiled MINR-057 registration gap.

## Problem

`upgrade_multisig` was a flat M-of-N over compile-time parameters:

- **#126 / MINR-054 / audit-3 finding 05** — a duplicated signer halved the threshold; threshold 0 authorised everything; threshold > n could never be met. A parameter is applied off-chain and appears in no transaction, so there was nowhere to guard it.
- **#127 / M7** — every rotation produced a new applied script, so a new credential and a full `upgrade_cred` handover (two transactions since #125).
- **MINR-057** — no `publish` handler. Registrable only via the legacy no-witness certificate, which the era after Conway withdraws; after that no new instance could ever be registered.

## Change

One validator over `(utxo_ref)`, one hash serving as NFT policy, config address and `upgrade_cred`:

| handler | role |
|---|---|
| `mint` | one-shot config NFT at its own address, datum = `MultisigScript` tree, `well_formed` |
| `spend` | rotation: old tree approves, new tree `well_formed`, NFT continues; `upgrade_cred` does not move |
| `withdraw` | trampoline target: config found by NFT among reference inputs, tree evaluated against signatories, validity range and withdrawals |
| `publish` | `RegisterCredential` only |

`MultisigScript` and `satisfied` are SundaeSwap's (`aicone`, Apache-2.0, `main@753436d6caea`), copied into `lib/multisig.ak` with their test; `satisfied_payload` omitted. Copied rather than depended on: the repo is unpublished and this project takes no on-chain dependency beyond stdlib and fuzz.

`well_formed`, enforced on `mint` and `spend`: 28-byte hashes; list nodes non-empty and duplicate-free (`AllOf []` is vacuously true); `1 ≤ required ≤ n`; tree ≤ `max_size`. Sundae's `AtLeast` has the M-of-N defects itself; the rail is what closes them.

**`max_size = 20`**, measured (worst case: flat `AtLeast`, every member signing; 14M mem budget):

| cap | withdraw / upgrade | update / rotation |
|---|---|---|
| 16 | 0.90M / 262M | 1.91M / 682M — 14% mem |
| **20** | **1.23M / 358M** | **2.65M / 970M — 19%** |
| 24 | 1.61M / 468M | 3.51M / 1.31B — 25% |
| 32 | 2.52M / 733M | 5.61M / 2.16B — 40% |

The cap is in the validator's bytes; changing it later is a handover.

Also, in both spend handlers (this and `protocol_params`): `expect_match(…, fn(_, _) { True })` → `expect_match_assets` (−2.27M cpu, +55 B per validator); and the "exactly one input at this address" rail removed — the singular continuing output, exact non-ADA conservation and the one-shot NFT already fail a second input's own run. A junk-run test in each file covers it and dies when the value rail is neutered.

Follow-up commit renames `has_policy_nft_strictly` → `expect_only_nft_of_policy` (four sites, byte-identical).

## Verification

573/573 in both environments (was 512). `lib/multisig.ak`: Sundae's `satisfying` + 16 unit + 8 property tests. Validator: 44 unit, plus two layout goldens pinning the seven constructor indices. All 23 rails mutation-verified against the full suite:

| library | killed | validator | killed |
|---|---|---|---|
| sig hash length | 2 | seed consumed | 1 |
| script hash length | 2 | one NFT minted | 2 |
| non-empty node | 3 | genesis well-formed | 2 |
| distinct children | 2 | genesis no ref script | 1 |
| `required ≥ 1` | 4 | genesis at own address | 2 |
| `required ≤ n` | 5 | datum required | 1 |
| size cap | 2 | singleton output | 1 |
| | | update no ref script | 1 |
| | | non-ADA conserved | 3 |
| | | new tree well-formed | 2 |
| | | old tree approves | 2 |
| | | script account only | 1 |
| | | config found by NFT | 1 |
| | | tree satisfied | 7 |
| | | registration only | 1 |

## Surface

Full table in `CONTRACT_SURFACE_CHANGES.md`. Parameters `(signers, threshold)` → `(utxo_ref)`; four new blueprint titles; new one-shot NFT `UpgradeMultisig` at `address.from_script(hash)` with no stake credential; new datum type. Deployment gains one genesis transaction and a `readFrom` of the config UTxO on every upgrade. A rotation and a protocol upgrade are two transactions. Narrative docs deferred to #131.


fabianbormann (approved):
LGTM!

PR #137, opened 2026-09-10
Make an AtLeast count mean distinct parties

nemo83:
Closes the multisig finding from the comment-versus-code triage. This is the one change in that pass that moves validator bytes.

**The defect.** `well_formed` bounded a tree's shape but not the evidence a satisfying path requires. Two shapes read as two parties and were satisfied by one signature:

```
AtLeast { required: 2, scripts: [AllOf { scripts: [alice] }, alice] }
AtLeast { required: 2, scripts: [alice, After { time: 0 }] }
```

The first survives structural de-duplication, because the two children are different values even though one signature satisfies both. The second pads the count with a leaf that everyone satisfies. Both were well-formed, both installable through mint and spend, and both then authorised an upgrade on alice alone. Separately, a tree of time leaves alone was also well-formed, which is a permissionless authority.

Neither is injectable by an outsider, since installing any tree requires satisfying the current one. What makes them worth fixing is that the code already spends bytes rejecting three defects with exactly the same reachability: an empty conjunction, a zero threshold, and a threshold above the child count. The threat model this validator already adopted does not distinguish those from these. And the upgrade credential governs the dispatcher credential, so it reaches token custody.

**The rule.** No installable tree may be satisfiable without producing a credential. And every child of an `AtLeast` must cost a credential of its own, with no credential appearing under two children. A key hash and a script hash are tagged apart, since one is met by a signature and the other by a withdrawal; conflating them was a real bug in the first draft of this rule, and a test now guards it.

It is deliberately sufficient rather than necessary. It also refuses overlapping but sound trees such as `AtLeast { 2, [AllOf [a, b], AllOf [b, c]] }`, which genuinely needs three signatures. Overlapping members in an n-of-m are ambiguous to read, and a rule decidable by looking at one node was judged worth the shapes it costs. Nested groups stay legal when disjoint, so "two of: the hot pair, the cold pair, the governance script" still works, and a time leaf is still legal where it constrains a party instead of standing in for one.

**A property test that could not fail.** `prop_no_evidence_never_satisfies` never called `well_formed`, so the rail was not on its execution path and no mutation of it could ever fail the test. It now draws trees that may be ill-formed, asserts the implication, and dies when either new rule is removed.

**Cost.** The first draft of the rule cost +54% memory and +41% cpu on the update path at the size cap, which would have made the cost table's justification for `max_size = 20` false: 29% of the mainnet memory budget against the 19% it claims. That is not acceptable for a number baked into the bytes, since changing the cap later is an authority handover.

A sorted-merge fold was tried and was worse than what it replaced, 5.86 M against 4.04 M, because rebuilding the accumulator at each merge costs more on a flat tree than the quadratic scan it removes. `list.sort` is insertion sort, so sorting does not help either.

What worked is a fast path. `children_ok` already proves an `AtLeast`'s members structurally distinct, and two distinct credential leaves are two distinct credentials, so when every member is a `Signature` or a `Script` the evidence walk is skipped entirely. That is the shape an authority writes and the shape the cap was measured against; nested members still pay.

| worst case at the cap | before | after |
|---|---|---|
| update, mem | 2.63 M | 2.84 M (+7.9%) |
| update, cpu | 963 M | 1.02 B (+6.2%) |
| withdraw | 1.23 M / 358 M | unchanged |

Two mutants pin the fast path as an optimisation rather than a second rule: taking it for every child kind kills five tests, and never taking it changes nothing at all.

**Blast radius.** `upgrade_multisig` grows from 2,791 to 3,395 bytes and takes a new hash. The other 29 validators are byte-identical to main, verified per validator. Nothing is parameterised by this script and it is parameterised by none, so nothing cascades. A live deployment would need one protocol-params datum edit, since `upgrade_cred` names this script's withdraw credential; there is no live deployment yet. The `withdraw` handler does not call `well_formed`, so authorising an upgrade costs exactly what it did before. Only the two config-write paths pay.

**Verification.** 615 tests pass under `aiken check -D` and under `--env with_assertions`; `aiken fmt --check` clean; the blueprint reproduces. Each rule was mutation-verified separately: removing the permissionless rule kills the time-only test and the property; removing the distinct-parties rule kills all three inflation tests; and each half of that rule, substituted independently, kills exactly the test named for it.


fabianbormann (approved):

nemo83:
Closing as won't fix unless it will turn required in the future

PR #144, opened 2026-09-23
integrate aiken v1.1.24, leveraging value builtin where possible.

KtorZ:
> [!IMPORTANT]
>
> Work-in-progress; the new compiler isn't even out _yet_ (but available on `main`). I've been using this project as an extra > sandbox to try out things. 

Some more simplifications can eventually also be achieved with aiken's [stdlib v4.0.0](https://github.com/aiken-lang/stdlib/pull/136)

## Early results

Script size reduced by ~550 bytes.

#### Before

```
many_tokens (projected max size = 1972)
many_inputs (projected max size = 339)
many_policies (projected max size = 232)
many_outputs (projected max size = 333)
```

#### After

```
many_tokens (projected max size = 12303)
many_inputs (projected max size = 797)
many_policies (projected max size = 281)
many_outputs (projected max size = 762)
```


validators/programmable_logic/transfer.ak
```diff
@@ -141,29 +147,14 @@ fn collect_input_assets(
...
-              _ -> Pair(i_policy, tokens.union(mint_at_policy, i_tokens))
-            },
-            ..apply_mint_to_known_policies(i_rest, mint_after)
-          ]
-        }
-      }
+  if value.zero == mint {
+    inputs
+  } else {
+    builtin.union_value(inputs, builtin.un_value_data(mint))
```
KtorZ: Might be more efficient to insert one by one than to convert+union; to be measured.

